PR #15, opened 2014-07-15
Rename gem to twingly-url

jage:
This is the start of a rewrite. It will become a more general gem, where normalization is just one of many libraries.
- [x] Rename `twingly-url-normalizer` to `twingl/url/normalizer`
- [x] Split up the parsing in`twingly/url/normalizer` to `twingly/url`
- [x] Add new methods to validate URLs (to be used in [Ganges](https://github.com/twingly/ganges) and [Norrstrom](https://github.com/twingly/norrstrom))
- [x] Add hashing functions for documentdb, blogstreamdb
- [x] Rename repository and gemspec

### DIFF
--- a/twingly-url.gemspec
+++ b/twingly-url.gemspec
@@ -8,7 +8,7 @@ Gem::Specification.new do |s|
   s.platform    = Gem::Platform::RUBY
   s.authors     = ["Johan Eckerström"]
   s.email       = ["johan.eckerstrom@twingly.com"]
-  s.homepage    = "http://github.com/twingly/twingly-url-normalizer"
+  s.homepage    = "http://github.com/twingly/twingly-url"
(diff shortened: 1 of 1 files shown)


test/unit/url_test.rb
```diff
@@ -0,0 +1,13 @@
+require 'test_helper'
+
+class UrlTest < Test::Unit::TestCase
+  context ".validate" do
+    should "return true for a valid url" do
+      assert Twingly::URL.validate("http://blog.twingly.com/"), "Should be valid"
+    end
+
+    should "return false for a valid url" do
```
dentarg: Fixed in https://github.com/twingly/twingly-url/commit/cce261155aeaba84994e54f543dc2dc7cbcfaeae

Issue #12, opened 2014-04-21
Discrepancy with .NET normalization


walro:
Using url: `http://www.blog.com/feed/` the normalizer returns `http://www.blog.com/feed/` while the .NET based normalization would return `http://www.blog.com/feed`. So this library is not safe to use in conjunction with .NET based projects, such as Autoping.

The .NET filtering is done in: `\trunk\Utilities\Filtering\UrlFilter.cs`

In: `\trunk\Utilities\Filter.Test\TestUrlFilter.cs` there are lots of tests defining the behavior (Including tests for stripping trailing slash in path).

roback:
This is needed to test whether a url is added to autoping, as im trying to do in https://trello.com/c/KG5vNOEA/1286-sommar2015-verktyg-for-att-jobba-med-url-listor

jage:
We discussed this on the sprint meeting for [sprint 32](https://trello.com/c/v1BCxbyH/1355-sprint32-planering-och-uppfoljning) and decided that this isn't critical. We would very much like this, but it won't be prioritized now.

PR #23, opened 2015-01-13
Downcase URLs in normalization

### LINKED ISSUE #8: Always return normalized urls in lower case
This would avoid problems with Autoping as it currently requires at least the protocol to be in lower-case. 

I can't think of any drawback of doing this, can anyone else?

jage:
URLs are not case-insensitive, but since our normalized URLs are just
used internally, it's ok. The original URL should always be used when
actually requesting a resource.

Close #8

### DIFF
--- a/lib/twingly/url/normalizer.rb
+++ b/lib/twingly/url/normalizer.rb
@@ -28,7 +28,7 @@ module Twingly
           result.url.path = "/"
         end
 
-        result.url.to_s
+        result.url.to_s.downcase
       end
     end
   end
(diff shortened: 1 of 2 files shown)


dentarg:
Hmm.

dentarg:
> Hmm.

I was thinking of the fact that Ruby `#downcase` [doesn't handle UTF-8](https://www.google.se/search?q=ruby+downcase+utf8).

    pry(main)> Twingly::URL::Normalizer.normalize("http://www.fooBARfooÅÄÖfoo.com").first.downcase
    => "http://www.foobarfooÅÄÖfoo.com/"

I don't think that's a stopper here, I just wanted to make sure it doesn't blow up.

jage:
Good thinking.

walro:
Hmm, I wonder what our .NET implementation does here actually. Not entirely sure it down-cases.... We need to take alook at this in #12 too.

I think we downcased here to solve an issue in BRM.

Issue #26, opened 2015-04-14
Shoulda-context does not seem to work with Ruby 2.2


walro:
I tried to change so that we test against 2.2. First I ran into to the following problem:

```ruby
/Users/robin/.gem/ruby/2.2.1/gems/shoulda-context-1.2.1/lib/shoulda/context/test_framework_detection.rb:9:in `block in possible_test_frameworks': cannot load such file -- test/unit (LoadError)
```

After adding `test-unit` as a development dependency: `s.add_development_dependency "test-unit", "~> 3"`:

```ruby
Error: test_: .extract_urls should detect two urls in an Array. (NormalizerTest)
: NameError: uninitialized class variable @@current in MiniTest::Unit::TestCase
/Users/robin/.gem/ruby/2.2.1/gems/minitest-4.7.5/lib/minitest/unit.rb:1297:in `current'
(eval):8:in `must_equal'
/Users/robin/Workspace/git/twingly-url/test/unit/normalization_test.rb:64:in `block (2 levels) in <class:NormalizerTest>'
/Users/robin/.gem/ruby/2.2.1/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `instance_exec'
/Users/robin/.gem/ruby/2.2.1/gems/shoulda-context-1.2.1/lib/shoulda/context/context.rb:413:in `block in create_test_from_should_hash'
```

Looking at https://github.com/thoughtbot/shoulda-context they only test against 2.0.0 so I think we might want to remove this dependency.

walro:
Fixed in #29

PR #29, opened 2015-04-30
Turn is unmaintained

### LINKED ISSUE #27: Turn is unmaintained
We can use minitest-reporters instead as in: https://github.com/twingly/stanley/pull/161/files

walro:
This PR does a few things

Move to minitest
Move to minitest-reporters which in turn move to non-built-in version of minitest
which in turn has deprecated/removed a few things

Fixes #27 and #26

### DIFF
--- a/test/profile/normalize_performance_test.rb
+++ b/test/profile/normalize_performance_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 require 'test_profile'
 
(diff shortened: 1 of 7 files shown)


walro:
Also probably fixes #26

walro:
> Also probably fixes #26

Yes, seems to work with ruby 2.2. I'll make another PR after this to ensure we test with 2.2

jage:
LGTM

PR #32, opened 2015-06-04
Make sure Digest loading is thread-safe

### LINKED ISSUE #20: Digest is not threadsafe
In twingly/zambezi#159 we encountered the error

    RuntimeError: Digest::Base cannot be directly inherited in Ruby

[Google](https://www.google.se/search?q="Digest::Base+cannot+be+directly+inherited+in+Ruby") led me to https://github.com/opscode/chef/issues/1793 which states:

> After some debugging we've pinpointed this error to Digest not being threadsafe.

jage:
I'm defining the hashers in constants to avoid any performance overhead
by the mutex since we hash a lot of URLs

Digest() is only thread safe in Ruby 2.2.0 and newer. Patch: https://github.com/ruby/ruby/commit/c02fa39463a0c6bf698b01bc610135604aca2ff4

I tried creating a test for this, but it's hard to provoke the error
with the old code to verify the test.

The patch version number needs a bump when this goes in.

Closes #20

### DIFF
--- a/lib/twingly/url/hasher.rb
+++ b/lib/twingly/url/hasher.rb
@@ -1,29 +1,34 @@
-require 'digest/md5'
-require 'digest/sha2'
+require 'digest'
 
 module Twingly
   module URL
     module Hasher
       module_function
 
+      # Instantiate digest classes in a thread-safe manner
+      # This is important since we don't know how people will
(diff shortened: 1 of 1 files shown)


walro:
LGTM

dentarg:
:+1:

Issue #28, opened 2015-04-28
normalize method can't handle URLs with punycoded TLD


walro:
```ruby
[12] pry(main)> Twingly::URL::Normalizer.normalize("http://xn--80aesdcplhhhb0k.xn--p1ai/")
=> []
```

Page loads fine in Chrome though, whois works fine too.

walro:
It was pointed out that the TLD is really funky here.

dentarg:
Strange, xn--p1ai is in [the public suffix list](https://github.com/weppos/publicsuffix-ruby/blob/1c452b92902d02c75dbc8ed4a8630ab94e461ad2/data/definitions.txt#L6741-L6743)

I tested too, with public_suffix 1.5.1, got the same as above.

Don't really know of public_suffix work, if that above should be enough to support xn--p1ai or if they are missing something. 

Could it an encoding issue?

walro:
> Could it be an encoding issue?

Yeah, to get it working (with public-suffix) we need to go from punycode back to utf-8:

```ruby
irb(main):005:0> PublicSuffix.valid?("xn--80aesdcplhhhb0k.xn--p1ai")
=> false
irb(main):006:0> PublicSuffix.valid?("domain.рф")
=> true
```

Public suffix won't add support for it: https://github.com/weppos/publicsuffix-ruby/issues/24

We could use https://github.com/mmriis/simpleidn (I found other, even less maintained, alternatives too) to do this ourselves.

jage:
We should analyze our data and how many punycode TLDs do we have.

Issue #18, opened 2014-10-30
Remove www. for domains with "subdomains"


jage:
`http://www.blog.twingly.com/` should be normalized to `http://blog.twingly.com/` according to ` /trunk/Utilities/Filtering/UrlFilter.cs` and `/trunk/Utilities/Filter.Test/TestUrlFilter.cs`.

roback:
We can't reproduce this. Proof in: http://systemutil.live.lkp.primelabs.se/documentdb/8155421525997345110 link still has www before subdomain.

Post: https://twfakeblog.herokuapp.com/urls/uris/2015/09/08/moar-url-test.html

jage:
:+1: 

Might have been fooled by WordPress that does http://www.blog.twingly.com/ -> http://blog.twingly.com/

```
$ http http://www.blog.twingly.com/
HTTP/1.1 301 Moved Permanently
Location: http://blog.twingly.com/
```

PR #37, opened 2015-09-08
Sync known behaviour with .NET

### LINKED ISSUE #12: Discrepancy with .NET normalization

### LINKED ISSUE #17: Should normalize IDN properly
twingly-url should normalize [internationalized domain names](https://en.wikipedia.org/wiki/Internationalized_domain_name) (IDN) properly.

```ruby
[13] pry(main)> Twingly::URL::Normalizer.normalize "http://www.åäö.se"
=> ["http://www.åäö.se/"]

[18] pry(main)> Addressable::URI.parse("http://www.åäö.se").normalize.to_s
=> "http://www.xn--4cab6c.se/"
```

Here is how the addressable gem does it:

* lib/addressable/uri.rb [`normalized_host`](https://github.com/sporkmonger/addressable/blob/2b9143ca02a0d83f90be3e7603fdfa7fa99bbb7d/lib/addressable/uri.rb#L1063-L1078)

* lib/addressable/idna/pure.rb [`to_ascii`](https://github.com/sporkmonger/addressable/blob/2b9143ca02a0d83f90be3e7603fdfa7fa99bbb7d/lib/addressable/idna/pure.rb#L64-L88)

roback:
Closes #12, #17, #18, #28.

### DIFF
--- a/spec/lib/twingly/url/normalization_spec.rb
+++ b/spec/lib/twingly/url/normalization_spec.rb
@@ -12,77 +12,43 @@ describe Twingly::URL::Normalizer do
       expect { normalizer.normalize([]) }.not_to raise_error
     end
 
-    it "handles URL with ] in it" do
-      url = "http://www.iwaseki.co.jp/cgi/yybbs/yybbs.cgi/%DEuropean]buy"
-      expect { normalizer.normalize(url) }.not_to raise_error
-    end
-
-    it "handles URL with reference to another URL in it" do
-      url = "http://news.google.com/news/url?sa=t&fd=R&usg=AFQjCNGc4A_sfGS6fMMqggiK_8h6yk2miw&url=http:%20%20%20//fansided.com/2013/08/02/nike-decides-to-drop-milwaukee-brewers-ryan-braun"
-      expect { normalizer.normalize(url) }.not_to raise_error
-    end
-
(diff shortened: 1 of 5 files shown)


spec/lib/twingly/url/normalization_spec.rb
```diff
@@ -32,6 +34,25 @@
...
+      url = "http://www.xn--4cab6c.se/"
+
+      expect(normalizer.normalize(url)).to eq([url])
+    end
+
+    it "handles URL with punycoded TLD" do
+      url = "http://www.example.xn--p1ai/"
+
+      expect(normalizer.normalize(url)).to eq([url])
+    end
```
dentarg: We should add test without subdomain here, I suspect it will fail... because of https://github.com/sporkmonger/addressable/issues/199

Issue #31, opened 2015-05-20
Rework exceptions


jage:
First of, we shouldn't raise that many exceptions if we can avoid it. I've made some comments on what we should return on exception here below:
- `Twingly::URL.parse` - Returns a Struct with `#url` and `#domain` accessors: _Return "empty" struct on exception_
- `Twingly::URL.validate` - Validates a URL: _Return false on exception_
- `Twingly::URL::Normalizer.normalize(string)` - Extracts URLs from string (Array): _Return empty array on exception_
- `Twingly::URL::Utilities.remove_scheme(url)` - Removes scheme from HTTP/HTTPS URLs (`http://twingly.com` -> `//twingly.com`): _Always run to_str on input, always remove scheme (if there's nothing to remove, return same input as string)_

Not sure about the hashers: _Run to_str on input (won't accept nil etc), always return hash?_
- `Twingly::URL::Hasher.taskdb_hash(url)` - MD5 hexdigest:
- `Twingly::URL::Hasher.blogstream_hash(url)` - MD5 hexdigest
- `Twingly::URL::Hasher.documentdb_hash(url)` - SHA256 unsigned long, native endian digest
- `Twingly::URL::Hasher.autopingdb_hash(url)` - SHA256 64-bit signed, native endian digest
- `Twingly::URL::Hasher.pingloggerdb_hash(url)` - SHA256 64-bit unsigned, native endian digest

But if we happen to leak an exception from the libraries used in this gem, they should be tagged by a `Twingly::URL` error module to make it easier on the user to handle.

``` Ruby
# Pseudo code
module Twingly::URL
  module Error
  end
end

# ...

def self.normalize_url
  # Do some fancy stuff
rescue Exception => error
  error.extend(Twingly::URL::Error)
  raise
end
```

jage:
As an alternative to module tagging we could wrap the exceptions, I like how Sequel does it:

* Rescue errors, pass to `raise_error` https://github.com/jeremyevans/sequel/blob/4e222141649eb610bf5bdfd59f117a98e364e8c5/lib/sequel/adapters/mysql2.rb#L72-L100
* `raise_error` converts exception class: https://github.com/jeremyevans/sequel/blob/9cc1ba7ecc682fe2c5af4c876356806f83ce0b02/lib/sequel/database/misc.rb#L419-L427
[...]

PR #42, opened 2015-10-13
Work with Twingly::URL objects instead of strings

### LINKED ISSUE #11: Capability for extracting origin part of an URL
Would be nice if the normalizer could extract the normalized host:

`http://domain.com.br/some_path` -> `http://www.domain.com.br/`

### LINKED ISSUE #35: Do not return nil
`url` and `domain` can be `nil`: https://github.com/twingly/twingly-url/blob/7ea8c42165580e47a0664906de070e62f9925a94/lib/twingly/url.rb#L29

That makes it harder to use this library, because you always need to check for `nil`... (example: https://github.com/twingly/norrstrom/issues/56)

### LINKED ISSUE #40: Return objects instead of strings
Ideas:

```ruby
Twingly::URL::Normalize.normalize(potential_urls)
# => [Twingly::URLObject]
```

```ruby
url = Twingly::URL.new(potential_url)
url.valid?
url.without_host.to_s
url.without_path.to_s
url.normalized.to_s

Twingly::URL.extract_urls(text).each do |url|
  puts url.without_host.to_s
end
```

### LINKED ISSUE #45: Ensure proper behaviour for edge-case input data
Ensure we handle the following input data like we do in .NET

* Whitespaces
* Newlines
* Garbage
* Non-URL

twingly-mob:
Related to #11, #31, #35, https://github.com/twingly/chapman/issues/69.

TODO:

- [x] #11 
- [x] `new`
- [x] Test name
- [x] #45 

Close #11
Close #35 
Close #40 
Close #45


jage:
I'd like to look at the `method_missing` implementation, should we implement `respond_to?` as well?
Would also like to add more tests with spaces and newlines around valid URLs.

jage:
I'd like a more dense inspect output.

```
$ bundle console
irb(main):001:0> Twingly::URL.parse("http://duh.se")
=> #<Twingly::URL:0x007fb3221e0fc8 @addressable_uri=#<Addressable::URI:0x3fd9910f0488 URI:http://duh.se>, @public_suffix_domain=#<PublicSuffix::Domain:0x007fb3222c5a88 @tld="se", @sld="duh", @trd=nil>>
```

jage:
Should this be true or false?

```Ruby
irb(main):007:0> url = "http://duh.se/"
=> "http://duh.se/"
irb(main):008:0> Twingly::URL.parse(url) == Twingly::URL.parse(url)
```

walro:
> Should this be true or false?

```ruby
[7] pry(main)> uri1 = URI.parse("http://duh.se")
=> #<URI::HTTP http://duh.se>
[8] pry(main)> uri2 = URI.parse("http://duh.se")
=> #<URI::HTTP http://duh.se>
[9] pry(main)> uri1 == uri2
=> true
```

I'd say true :)

jage:
http://ruby-doc.org/core-2.2.3/Comparable.html might be interesting to use.

I'm thinking:

```Ruby
include Comparable

def <=>(anotherUrl)
  self.to_s <=> anotherUrl.to_s
end
```

twingly-mob:
What can we do about this:

```ruby
irb(main):010:0> Twingly::URL.new.domain
NoMethodError: undefined method `domain' for nil:NilClass
	from /Users/twingly/repos/twingly-url/lib/twingly/url.rb:42:in `domain'
```

dentarg:
Actually, we haven't solved #11 

> Would be nice if the normalizer could extract the normalized host:

>`http://domain.com.br/some_path` -> `http://www.domain.com.br/` 

I think the original title of that issue confused us, "Capability for extracting just the host part of an URL", we have `#host` but that [doesn't include the scheme](https://github.com/twingly/twingly-url/blob/8975fd870e3866ec3d9ca6f865a361e296c7ef45/spec/lib/twingly/url_spec.rb#L73-L76).

What we want is called "origin" in Addressable. We want to use it in Norrstrom [ping_message.rb](https://github.com/twingly/norrstrom/blob/8988f1a0bbb78b5ced8780e6a1b65bd1505421df/lib/ping_message.rb) ([specs](https://github.com/twingly/norrstrom/blob/8988f1a0bbb78b5ced8780e6a1b65bd1505421df/spec/ping_message_spec.rb) + [examples](https://github.com/twingly/norrstrom/blob/8988f1a0bbb78b5ced8780e6a1b65bd1505421df/spec/examples.rb#L107-L135)).

dentarg:
> We want to use it in Norrstrom

We don't want to use normalization there

walro:
> new

I am thinking we implement `initialize` and let it accept `Addressable::URI` and `PublicSuffix::Domain`. Rename `.setup` to `.internal_parse`. I opted to make it public so we wouldn't have to `#send`, added the `internal` prefix instead to denote that you shouldn't use this in clients.

WDYT @twingly/dev ?

```ruby
index a31fc66..2924bff 100644
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -16,12 +16,41 @@ module Twingly
     def self.parse(potential_url)
       potential_url = String(potential_url)
 
-      # TODO: Can we make this less send-y?
-      self.new.send(:setup, potential_url)
+      internal_parse(potential_url)
     rescue Twingly::URL::Error, Twingly::URL::Error::ParseError => error
       NullURL.new
     end
 
+    def self.internal_parse(potential_url)
+      if potential_url.is_a?(Addressable::URI)
+        addressable_uri = potential_url
+      else
+        addressable_uri = Addressable::URI.heuristic_parse(potential_url)
+      end
+
+      raise Twingly::Error::ParseError if addressable_uri.nil?
+
+      public_suffix_domain = PublicSuffix.parse(addressable_uri.display_uri.host)
+
+      self.new(addressable_uri, public_suffix_domain)
+    rescue Addressable::URI::InvalidURIError, PublicSuffix::DomainInvalid => error
+      error.extend(Twingly::URL::Error)
+      raise
+    end
+
+    def initialize(addressable_uri, public_suffix_domain)
+      unless addressable_uri.is_a?(Addressable::URI)
+        raise ArgumentError, "First parameter must be an Addressable::URI"
+      end
+
+      unless public_suffix_domain.is_a?(PublicSuffix::Domain)
+        raise ArgumentError, "Second parameter must be a PublicSuffix::Domain"
+      end
+
+      @addressable_uri      = addressable_uri
+      @public_suffix_domain = public_suffix_domain
+    end
+
     def scheme
       addressable_uri.scheme
     end
@@ -65,7 +94,7 @@ module Twingly
       normalized_url.host   = normalized_host
       normalized_url.path   = normalized_path
 
-      setup(normalized_url)
+      self.class.internal_parse(normalized_url)
     end
 
     def normalized_scheme
@@ -107,23 +136,6 @@ module Twingly
 
     attr_reader :addressable_uri, :public_suffix_domain
 
-    def setup(potential_url)
-      if potential_url.is_a?(Addressable::URI)
-        @addressable_uri = potential_url
-      else
-        @addressable_uri = Addressable::URI.heuristic_parse(potential_url)
-      end
-
-      raise Twingly::Error::ParseError if addressable_uri.nil?
-
-      @public_suffix_domain = PublicSuffix.parse(addressable_uri.display_uri.host)
-
-      self
-    rescue Addressable::URI::InvalidURIError, PublicSuffix::DomainInvalid => error
-      error.extend(Twingly::URL::Error)
-      raise
-    end
-
     def normalize_blogspot(host, domain)
       if domain.sld.downcase == "blogspot"
         host.sub(/\Awww\./i, "").sub(/#{domain.tld}\z/i, "com")
index d654330..13c4e83 100644
@@ -45,6 +45,14 @@ describe Twingly::URL do
     end
   end
 
+  describe "#initialize" do
+    context "when given input parameters of wrong types" do
+      it "raises an error" do
+        expect { described_class.new("a", "b") }.to raise_error
+      end
+    end
+  end
+
   describe "#scheme" do
     subject { url.scheme }
     it { is_expected.to eq("http") }
```

dentarg:
Sounds good to me

roback:
:+1: Better than send :)

jage:
:+1:

walro:
Are we done?

lib/twingly/url/null_url.rb
```diff
@@ -0,0 +1,30 @@
+module Twingly
+  class URL
+    class NullURL
+    include Comparable
```
dentarg: indentation missing here?

walro: indeed

dentarg:
This could be bad

```diff
$ git d
diff --git a/spec/lib/twingly/url_spec.rb b/spec/lib/twingly/url_spec.rb
index 13c4e83..b54c467 100644
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -128,6 +128,17 @@ describe Twingly::URL do
     end
   end

+  describe "#normalized.origin" do
+    subject { described_class.parse(url).normalized.origin }
+
+    context "adds a trailing slash if missing in origin" do
+      let(:url)      { "http://twingly.com/" }
+      let(:expected) { "http://www.twingly.com/" }
+
+      it { is_expected.to eq(expected) }
+    end
+  end
+
   describe "#normalized" do
     subject { described_class.parse(url).normalized.to_s }
```

walro:
> This could be bad

As per discussion IRL, I don't think "/" should be part of any origin, normalized or not. We should view trailing slash as "we always require a starting / in our paths".

spec/lib/twingly/url_spec.rb
```diff
@@ -0,0 +1,455 @@
...
+      it { is_expected.to eq(url) }
+    end
+
+    context "keeps www if the host already has it" do
+      let(:url) { "http://www.twingly.com/" }
+
+      it { is_expected.to eq(url) }
+    end
+
+    context "adds a trailing slash if missing in origin" do
```
dentarg: `/` is not part of origin (which is scheme, host and port), so the wording is wrong here

spec/lib/twingly/url_spec.rb
```diff
@@ -0,0 +1,455 @@
...
+      let(:url)      { "http://www.twingly.com" }
+      let(:expected) { "http://www.twingly.com/" }
+
+      it { is_expected.to eq(expected) }
+    end
+
+    context "ensures single trailing slash in origin" do
```
dentarg: `/` is not part of origin, so the wording is wrong here

spec/lib/twingly/url_spec.rb
```diff
@@ -0,0 +1,455 @@
...
+      let(:url)      { "http://www.twingly.com//" }
+      let(:expected) { "http://www.twingly.com/" }
+
+      it { is_expected.to eq(expected) }
+    end
+
+    context "removes trailing slash from path" do
```
dentarg: "from end of path unless path becomes empty"? :)

dentarg:
> As per discussion IRL, I don't think "/" should be part of any origin, normalized or not. We should view trailing slash as "we always require a starting / in our paths".

:+1:

twingly-mob:
@twingly/dev What about now?

spec/lib/twingly/url_spec.rb
dentarg: no need to do something here?

jage:
> @twingly/dev What about now?

I'm a bit worried about encodings, otherwise this looks great! (Could be handled in separate issue)

spec/lib/twingly/url_spec.rb
twingly-mob: We missed this, thanks. We'll change to your suggestion as we can't come up with something better. :)

PR #59, opened 2015-10-30
We support 2.2.x

### LINKED ISSUE #53: Drop support for older Ruby versions
< 2.2

walro:
Do we need to release a new major version because of this?

Close #53.


jage:
> Do we need to release a new major version because of this?

That's probably the most polite way of doing this. I vote for major version bump.

walro:
> That's probably the most polite way of doing this. I vote for major version bump.

Yeah, that's what I thought too.

dentarg:
Let's bring this into master, as we are now working on the next major.

Issue #60, opened 2015-10-30
Make more methods private


dentarg:
I tought of this when I saw [.to_addressable_uri](https://github.com/twingly/twingly-url/commit/25460f201c52b0beb60f55fdc144f03ba7bdd1f7). I think we want to decrease the number of methods we expose, so we know what we can change without breaking stuff for users.

We can do it like

```ruby
private_class_method \
def self.foo
end
```

And if we feel the need to test, there is https://github.com/barelyknown/rspec-context-private

walro:
> I tought of this when I saw .to_addressable_uri. I think we want to decrease the number of methods we expose, so we know what we can change without breaking stuff for users.

Hmm yeah, but sometimes you want to test your helper method, making them private makes that trickier.

twingly-mob:
#58 will make .to_addressable_uri private, but there are more.

We would like to make the following methods private

* Twingly::URL.internal_parse
* Twingly::URL.new
* Twingly::URL#initialize

dentarg:
We should also make our constants private

dentarg:
What we have been looking for: http://stackoverflow.com/questions/4950454/class-static-instance-initializers-i-e-factory-methods-in-ruby/4951715#4951715 (credit to @walro for the find)

dentarg:
Pretty sure there nothing left to do here and that this should have been closed.

PR #69, opened 2015-12-11
Make things private

dentarg:
- [x] `#new` and `.internal_parse`
- [x] Constants

Close to #60.

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -19,6 +19,8 @@ module Twingly
       IDN::Idna::IdnaError,
     ]
 
+    private_constant :ACCEPTED_SCHEMES, :ENDS_WITH_SLASH, :ERRORS
+
     class << self
       def parse(potential_url)
         potential_url = String(potential_url)
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -19,41 +19,37 @@ module Twingly
       IDN::Idna::IdnaError,
     ]
 
-    def self.parse(potential_url)
-      potential_url = String(potential_url)
-      potential_url = potential_url.scrub
-      potential_url = potential_url.strip
+    class << self
+      def parse(potential_url)
+        potential_url = String(potential_url)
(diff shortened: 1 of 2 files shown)


dentarg:
`rake profile:normalize` reports `Total: 1.791179` for `master` and `Total: 2.063033` for this branch

dentarg:
screen shot 2015-12-11 at 20 11 56

walro:
> rake profile:normalize reports Total: 1.791179 for master and Total: 2.063033 for this branch

I am guessing due to the extra three calls in `.parse` compared to `.internal_parse`. I think it's fine though.

LGTM!

walro:
> I am guessing due to the extra three calls in .parse compared to .internal_parse. I think it's fine though.

Ah, no, I think we make an extra parsing now, I think we should fix that.

lib/twingly/url.rb
```diff
@@ -101,7 +99,7 @@ def normalized
       normalized_url.host   = normalized_host
       normalized_url.path   = normalized_path
 
-      self.class.internal_parse(normalized_url)
+      self.class.parse(normalized_url)
```
walro: This will convert an addressable uri to a string and then back to a addressable uri. I think we want to avoid that.

walro: I think it's pretty easily amended by splitting `internal_parse` into two different ones, one for addressable and one for public suffix

walro: > I think it's pretty easily amended by splitting internal_parse into two different ones, one for addressable and one for public suffix

We already had a helper method apparently, I thought the change in 219cf37a3d53dfcd7a1c70e7560c508940eec4d6 made sense

walro:
Would be interesting to re-run the benchmarks now

dentarg:
> Would be interesting to re-run the benchmarks now

Total: 1.793219

walro:
Cool

dentarg:
@walro are we done? :)

walro:
I think so!

PR #75, opened 2016-01-12
The gem should load the version constant

dentarg:
Without this

    $ pry -r ./lib/twingly/url.rb
    [1] pry(main)> Twingly::URL::VERSION
    NameError: uninitialized constant Twingly::URL::VERSION
    from (pry):1:in `__pry__'

With this

    $ pry -r ./lib/twingly/url.rb
    [1] pry(main)> Twingly::URL::VERSION
    => "3.0.2"

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -4,6 +4,7 @@ require "public_suffix"
 
 require_relative "url/null_url"
 require_relative "url/error"
+require_relative "version"
 
 PublicSuffix::List.private_domains = false
 
--- a/lib/twingly/url/hasher.rb
+++ b/lib/twingly/url/hasher.rb
@@ -1,5 +1,7 @@
 require 'digest'
 
+require_relative "../url"
+
 module Twingly
   class URL
     module Hasher


walro:
Not sure why it "should", I noticed addressable does this, was that the inspiration or is it a best practice of sorts?

dentarg:
I just expect it to be available, I guess almost all gems have it exposed

dentarg:
Example from the Python world:


```python
$ python
Python 2.7.11 (default, Dec  5 2015, 14:44:53)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests
>>> requests.__version__
'2.0.1'
```

jage:
I like this.

walro:
> I just expect it to be available, I guess almost all gems have it exposed

Yes, but won't our hasher class suffer from this same problem still?

dentarg:
> Yes, but won't our hasher class suffer from this same problem still?

Yes, I didn't think of hasher

hasher should also load it

PR #78, opened 2016-01-29
No part of a URL should be nil

### LINKED ISSUE #52: domain, sld, trd (maybe others) can be nil
This breaks our applications.

```ruby
[8] pry(main)> Twingly::URL.parse("https://.com/").domain
=> nil
[9] pry(main)> Twingly::URL.parse("https://.com/").sld
=> nil
[10] pry(main)> Twingly::URL.parse("https://.com/").trd
=> nil
```

roback:
An attempt to fix #52

* URLs without `sld` now results in a `NullURL`. (Im not sure whether my assumption here is correct, but it works for the URLs given as examples in #52)
* Added `.to_s` to `#trd`

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -39,7 +39,7 @@ module Twingly
         raise Twingly::URL::Error::ParseError if public_suffix_domain.nil?
 
         raise Twingly::URL::Error::ParseError unless public_suffix_domain.sld
-
+        puts addressable_uri.origin
         new(addressable_uri, public_suffix_domain)
       rescue *ERRORS => error
         error.extend(Twingly::URL::Error)
@@ -67,35 +67,35 @@ module Twingly
     end
 
     def scheme
-      addressable_uri.scheme
+      addressable_uri.scheme.to_s
     end
 
     def trd
-      public_suffix_domain.trd
+      public_suffix_domain.trd.to_s
     end
 
     def sld
-      public_suffix_domain.sld
+      public_suffix_domain.sld.to_s
     end
 
     def tld
-      public_suffix_domain.tld
+      public_suffix_domain.tld.to_s
     end
 
     def domain
-      public_suffix_domain.domain
+      public_suffix_domain.domain.to_s
     end
 
     def host
-      addressable_uri.host
+      addressable_uri.host.to_s
     end
 
     def origin
-      addressable_uri.origin
+      addressable_uri.origin.to_s
     end
 
     def path
-      addressable_uri.path
+      addressable_uri.path.to_s
     end
 
     def without_scheme
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -39,7 +39,7 @@ module Twingly
         raise Twingly::URL::Error::ParseError if public_suffix_domain.nil?
 
         raise Twingly::URL::Error::ParseError unless public_suffix_domain.sld
-        puts addressable_uri.origin
+
         new(addressable_uri, public_suffix_domain)
       rescue *ERRORS => error
         error.extend(Twingly::URL::Error)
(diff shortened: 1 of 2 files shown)


lib/twingly/url.rb
```diff
@@ -38,6 +38,8 @@ def internal_parse(potential_url)
         public_suffix_domain = PublicSuffix.parse(addressable_uri.display_uri.host)
         raise Twingly::URL::Error::ParseError if public_suffix_domain.nil?
 
+        raise Twingly::URL::Error::ParseError unless public_suffix_domain.sld
```
walro: Follow suit with the if just above (change to if nil). :)

walro:
> URLs without sld now results in a NullURL. (Im not sure whether my assumption here is correct, but it works for the URLs given as examples in #52)

I think I like this part.

> Added .to_s to all getter methods (#scheme, #trd, etc.)

Not so sure about this part, one interpretation of nil is "it doesn't exist", where `""` could be more like, it exists but it happens to be empty.

lib/twingly/url.rb
roback: :+1: Didn't notice that

roback:
> Not so sure about this part, one interpretation of nil is "it doesn't exist", where "" could be more like, it exists but it happens to be empty.

My thought was that since `NullURL` returns an empty string for all methods, so should `URL`.

walro:
> My thought was that since NullURL returns an empty string for all methods, so should URL.

Yeah, and it might make sense here too. I think the calls to `#to_s` is what bothers me the most, in `NullURL`we hide it :)

roback:
> I think the calls to #to_s is what bothers me the most

I don't like them either, but at least its clear what is returned :)

walro:
> Added .to_s to all getter methods (#scheme, #trd, etc.)

Does any of the urls added in the `url_spec` exercise this? Just looking the code it seems like the check for `sld` would pick them all up.

roback:
> > Added .to_s to all getter methods (#scheme, #trd, etc.)

> Does any of the urls added in the url_spec exercise this? Just looking the code it seems like the check for sld would pick them all up.

No. I'm not sure how to test those methods with invalid urls since a `NullURL` is returned instead. I added a test for `#trd` though since that could be nil.

`to_s` isn't necessary on all methods, but now we have control over what gets returned instead of relying on public_suffix and addressable. (public_suffix returns `nil` for missing url components (`sld`, `trd`etc) while Addressable seems to prefer empty strings.)

dentarg:
:+1:

roback:
Should this be released as a patch version? I think it should, but the `.to_s` that was added to `#trd` is a breaking change, at least for norrstrom.

dentarg:
I think we are already in for a new major, with the other changes that have been made in master (the private methods stuff).

jage:
> Should this be released as a patch version? I think it should, but the .to_s that was added to #trd is a breaking change, at least for norrstrom.

Agree it _should_ be a patch, but since we know it's a breaking change I think it's better to bump the major version. We didn't specify in any documentation that `#trd` should return empty string, so basically we changed the behavior now. It's hard to call this a "bug" to the users of the gem. :-)

jage:
> Agree it should be a patch, but since we know it's a breaking change I think it's better to bump the major version. We didn't specify in any documentation that #trd should return empty string, so basically we changed the behavior now. It's hard to call this a "bug" to the users of the gem. :-)

If we first returned `""`, then by accident released a version that returned `nil`, this would have been a patch release IMO.

roback:
> I think we are already in for a new major, with the other changes that have been made in master (the private methods stuff).

Wasn't aware of that.

> We didn't specify in any documentation that #trd should return empty string, so basically we changed the behavior now.

Thats true, a new major version it is then :)

Issue #62, opened 2015-11-04
Bug in normalized_host in Addressable (ArgumentError: invalid byte sequence in UTF-8)


walro:
When 

``` ruby
irb(main):021:0> Twingly::URL.parse("http://some_site.net%C2")
ArgumentError: invalid byte sequence in UTF-8
    from /Users/robin/.gem/ruby/2.2.3/gems/addressable-2.3.8/lib/addressable/uri.rb:1097:in `host='
    from /Users/robin/.gem/ruby/2.2.3/gems/addressable-2.3.8/lib/addressable/uri.rb:2050:in `display_uri'
    from /Users/robin/Workspace/twingly/twingly-url/lib/twingly/url.rb:30:in `internal_parse'
    from /Users/robin/Workspace/twingly/twingly-url/lib/twingly/url.rb:21:in `parse'
    from (irb):21
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/cli/console.rb:14:in `run'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/cli.rb:308:in `console'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/cli.rb:10:in `start'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/bin/bundle:20:in `block in <top (required)>'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/lib/bundler/friendly_errors.rb:7:in `with_friendly_errors'
    from /Users/robin/.gem/ruby/2.2.3/gems/bundler-1.10.6/bin/bundle:18:in `<top (required)>'
[...]

PR #79, opened 2016-02-01
Temporary fix for UTF-8 bug in addressable

roback:
Temporary fix for #62.

Related to sporkmonger/addressable#224

Used the code in https://github.com/twingly/twingly-url/issues/62#issuecomment-176830731 as inspiration, but I had to modify it so we didn't have to call `addressable.display_uri` twice.

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -35,7 +35,9 @@ module Twingly
         scheme = addressable_uri.scheme
         raise Twingly::URL::Error::ParseError unless scheme =~ ACCEPTED_SCHEMES
 
-        public_suffix_domain = PublicSuffix.parse(addressable_uri.display_uri.host)
+        display_uri = addressable_display_uri(addressable_uri)
+
+        public_suffix_domain = PublicSuffix.parse(display_uri.host)
         raise Twingly::URL::Error::ParseError if public_suffix_domain.nil?
 
         new(addressable_uri, public_suffix_domain)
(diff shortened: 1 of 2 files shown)


walro:
I think this looks good

lib/twingly/url.rb
```diff
@@ -56,7 +58,19 @@ def to_addressable_uri(potential_url)
...
+        addressable_uri.display_uri
+      rescue ArgumentError => error
+        if error.message.include?("invalid byte sequence in UTF-8")
+          raise Twingly::URL::Error::ParseError
+        end
+
+        raise
+      end
+
+      private :new, :internal_parse, :to_addressable_uri, :addressable_display_uri
```
dentarg: what about splitting this line to multiple lines, so next time we get the nice diff :)

roback: Sure. What is the best way of doing that?

```ruby
private :new,
  :internal_parse,
  :to_addressable_uri,
  :addressable_display_uri

# or

private :new
private :internal_parse
private :to_addressable_uri
private :addressable_display_uri
```

For the second one we wouldn't have to add a `,` on the last line when adding a symbol, so I think that one looks best :).

jage: I like the second one (easier to remove/add line).

walro: I think the second one too

dentarg: Yeah, sure let's go with the second one :)

roback: e0fdb738aa09a4768a4ec95d335238cadc43eda7

roback:
I'll create a new gem version after #78 is merged too.

PR #86, opened 2016-08-30
Add example usage to README

dentarg:


jage:
How about using a `com.br`-domain just to illustrate `.tld` better?

dentarg:
@jage Sure

jage:
Maybe a note that we should check if the README example needs to be updated under "Release workflow"? Otherwise LGTM!

PR #88, opened 2016-08-30
Add Twingly::URL#ttld, "true TLD" getter

dentarg:
Many[1] ccTLDs have introduced a second level underneath their ccTLD, but
often, we don't care about the second level. We want the "true TLD".

Discussed in https://github.com/twingly/wally/issues/49.

[1]: https://en.wikipedia.org/wiki/Second-level_domain


spec/lib/twingly/url/null_url_spec.rb
```diff
@@ -35,6 +35,11 @@
     it { is_expected.to eq("") }
   end
 
+  describe "#ttld" do
+    subject { url.tld }
```
walro: tld -> ttld

dentarg:
Could `#true_tld` be a better name?

jage:
> Could #true_tld be a better name?

Don't think it helps, still an ambiguous name that you need to "learn". Might be easier to differentiate from `tld`, but you still need to read carefully since `sld`, `tld`, `trd` are also similar.

dentarg:
Sure, we can keep it as `#ttld`.

lib/twingly/url.rb
```diff
@@ -100,6 +100,10 @@ def tld
       public_suffix_domain.tld
     end
 
+    def ttld
```
jage: Maybe a comment why we call it `ttld`?

jage:
Just https://github.com/twingly/twingly-url/pull/88#discussion_r76943030, otherwise this LGTM!

lib/twingly/url.rb
dentarg: What would the comment say? Wouldn't it be better to call the method `true_tld` then?

dentarg: @jage Like this?

````ruby
    # Many ccTLDs have a second level[1] underneath their ccTLD, use this when
    # you don't care about the second level.
    # 
    # [1]: https://en.wikipedia.org/wiki/Second-level_domain
    def ttld
      tld.split(".").last
    end
````

Comment (jage) [inline lib/twingly/url.rb]: > What would the comment say? Wouldn't it be better to call the method true_tld then?

Nah, I don't think true is any better. "True" doesn't add any context when to use this.

> @jage Like this?

Yes, that's better.

File: spec/lib/twingly/url_spec.rb [hunk 3]
```diff
@@ -171,6 +171,11 @@ def valid_urls
     it { is_expected.to eq("co.uk") }
   end
 
+  describe "#ttld" do
+    subject { url.ttld }
+    it { is_expected.to eq("uk") }
```
walro: Maybe me being a 🐔  but I think I'd like a test for `#ttld` when we have a normal domain as well, like a simple `.com`. I am thinking to expose problems in the way we split the `#tld`.

dentarg: Sure

dentarg: 910373d

walro:
LGTM!

Issue #71, opened 2015-12-14
Feature: Expose Addressable's #normalized_host


dentarg:
Related to https://github.com/twingly/klondike/issues/31

Would be useful to expose [`#normalized_host`](http://www.rubydoc.info/gems/addressable/Addressable/URI#normalized_host-instance_method) from Addressable, because Ruby DNS libraries ([stdlib](https://bugs.ruby-lang.org/issues/4270), https://github.com/alexdalitz/dnsruby/issues/94) can't handle IDN very well (not at all I mean).

```ruby
[18] pry(main)> Dnsruby::DNS.new.getaddress(Addressable::URI.heuristic_parse("räksmörgås.josefßon.org").normalized_host)
=> #<Dnsruby::IPv4 155.4.17.102>

[19] pry(main)> Resolv.getaddress(Addressable::URI.heuristic_parse("räksmörgås.josefßon.org").normalized_host)
=> "155.4.17.102"
```

dentarg:
And because our own [`#normalized_host`](https://github.com/twingly/twingly-url/blob/v3.0.2/lib/twingly/url.rb#L111-L122) isn't at all suitable to use (we can break URLs). The terminology here is unfortunate...

walro:
Don't we really want a "to punycode" method somewhere?

jage:
> Don't we really want a "to punycode" method somewhere?

That would be better. I'm guessing it would be nice to get both `host` and all strings that contain `host` in both `ascii` and `utf8`.

dentarg:
I'm not really following... 

Is it something like this what we mean?

```ruby
Twingly::URL.parse("http://räksmörgås.josefßon.org/foobar").to_punycode
# => "http://xn--rksmrgs-5wao1o.josefsson.org/foobar"
```

How should we implement it? Should we use http://www.rubydoc.info/gems/addressable/Addressable/URI#normalized_host-instance_method or not? In my mind that's the most straight forward and lowest cost thing to do

Please elaborate your thoughts! :)

dentarg:
I haven't read everything at https://en.wikipedia.org/wiki/Punycode, but in my mind we only care about Punycode in the context of DNS, the host that is.

Maybe we want a method called `punycoded_host`?

jage:
> Is it something like this what we mean?

Yes.

> I haven't read everything at https://en.wikipedia.org/wiki/Punycode, but in my mind we only care about Punycode in the context of DNS, the host that is.

[...]

PR #90, opened 2016-09-09
Ensure normalized IDNA domains return ASCII strings

### LINKED ISSUE #85: Use PublicSuffix 2.0
This summer, PublicSuffix 2.0 was released. It sure looks interesting.

https://github.com/weppos/publicsuffix-ruby/blob/master/2.0-Upgrade.md ([current permalink](https://github.com/weppos/publicsuffix-ruby/blob/a08e80696600ca51e566c5af541ad4b9a1f72265/2.0-Upgrade.md))

> PublicSuffix 2.0 contains a rewritten internal representation and comparison logic, that drastically increases the lookup performance. The new version also changes several internal and external API.

### LINKED ISSUE #89: NormalizedURL#to_s returns punycode, other instance methods does not
Calling `#.to_s` or `#host` on a normalized URL returns a punycoded string.

Calling `#domain`, `#tld`, `#sld` etc. on the normalized url returns the non-punycoded version of the URL. What those methods have in common is that they all uses public_suffix.

```ruby
[3] pry(main)> url = Twingly::URL.parse("http://test.भारत")
=> #<Twingly::URL:0x3fe40956dc7c http://test.भारत>
[4] pry(main)> normalized_url = url.normalized
=> #<Twingly::URL:0x3fe40acdbfa8 http://www.test.xn--h2brj9c/>
[5] pry(main)> url.to_s
=> "http://test.भारत"
[6] pry(main)> normalized_url.to_s
=> "http://www.test.xn--h2brj9c/"
[7] pry(main)> normalized_url.tld
=> "भारत"
[8] pry(main)> normalized_url.domain
=> "test.भारत"
```

dentarg:
This PR ensure correct behaviour for normalized URLs (fixes #89), but it also changes the behaviour for non- normalized URL objects, see these snippets ([original comment](https://github.com/twingly/twingly-url/pull/90#issuecomment-247021829)):

Previous version (4.2.0):

```Ruby
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3fd876a72dd8 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3fd8767d277c https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)>
[4] pry(main)> u1.to_s
=> "https://www.foo.ایران.ir/bar"
[5] pry(main)> u2.to_s
=> "https://www.foo.xn--mgba3a4f16a.ir/bar"
[6] pry(main)>
[7] pry(main)> u1.domain
=> "foo.ایران.ir"
[8] pry(main)> u2.domain
=> "foo.ایران.ir"
[9] pry(main)>
[10] pry(main)> u1.tld
=> "ایران.ir"
[11] pry(main)> u2.tld
=> "ایران.ir"
```

This version:

```Ruby
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3fc621fd5454 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3fc621fb855c https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)>
[4] pry(main)> u1.to_s
=> "https://www.foo.ایران.ir/bar"
[5] pry(main)> u2.to_s
=> "https://www.foo.xn--mgba3a4f16a.ir/bar"
[6] pry(main)>
[7] pry(main)> u1.domain
=> "foo.ایران.ir"
[8] pry(main)> u2.domain
=> "foo.xn--mgba3a4f16a.ir"
[9] pry(main)>
[10] pry(main)> u1.tld
=> "ایران.ir"
[11] pry(main)> u2.tld
=> "xn--mgba3a4f16a.ir"
```

---

_Original description_

~~So, yes, this increases the load time quite a bit... :)~~ Not true anymore (see 2bd6447) Haven't run the profiler yet, but we could do that.

I'm open to suggestions about how we should deal with the extending of the default list, I just did something to get the discussion started.

---

Related to #71. Close #85 and #89.


dentarg:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;nEv13hUn&#x2F;1832-sprint40-feature-location-by-internationalized-country-code-top-level-domains"> Sprint40: FEATURE Location by Internationalized country code top-level domains</a>

roback:
> I'm open to suggestions about how we should deal with the extending of the default list, I just did something to get the discussion started.

We could also create a new list instead of changing the default one:

```ruby
    CUSTOM_SUFFIX_LIST = PublicSuffix::List.
      parse(File.read(PublicSuffix::List::DEFAULT_LIST_PATH), private_domains: false).
      tap do |list|
        list.indexes.keys.
        map    { |name| Addressable::IDNA.to_ascii(name) }.
        select { |name| name =~ /xn\-\-/ }.
        each   { |name| list << PublicSuffix::Rule.factory(name) }
      end

# ...

public_suffix_domain = PublicSuffix.parse(host, default_rule: nil, list: CUSTOM_SUFFIX_LIST)
```

lib/twingly/url.rb
```diff
@@ -6,7 +6,12 @@
 require_relative "url/error"
 require_relative "version"
 
-PublicSuffix::List.private_domains = false
+PublicSuffix::List.default = PublicSuffix::List.parse(
+  File.read(PublicSuffix::List::DEFAULT_LIST_PATH), private_domains: false)
+PublicSuffix::List.default.indexes.keys.
+  map { |name| Addressable::IDNA.to_ascii(name) }.
+  select { |name| name =~ /xn\-\-/ }.
```
walro: maybe we should add `^`, I think it's always in the start of the host?

roback:
This doesn't look right:

```ruby
[1] pry(main)> Twingly::URL.parse("https://www.foo.ایران.ir/bar").tld
=> "ایران.ir"
[2] pry(main)> Twingly::URL.parse("https://www.foo.ایران.ir/bar").normalized.tld
=> "ir"
```

roback:
> This doesn't look right:

Made a few changes that fixes things:

```diff
diff --git a/lib/twingly/url.rb b/lib/twingly/url.rb
index 7e7fef6..46ac6ca 100644
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -8,10 +8,12 @@ require_relative "version"

 PublicSuffix::List.default = PublicSuffix::List.parse(
   File.read(PublicSuffix::List::DEFAULT_LIST_PATH), private_domains: false)
-PublicSuffix::List.default.indexes.keys.
-  map { |name| Addressable::IDNA.to_ascii(name) }.
+PublicSuffix::List.default.each.
+  map { |rule| Addressable::IDNA.to_ascii(rule.value) }.
   select { |name| name =~ /xn\-\-/ }.
-  each { |name| PublicSuffix::List.default << PublicSuffix::Rule.factory(name) }
+  each { |name| PublicSuffix::List.default.add(PublicSuffix::Rule.factory(name), reindex: false) }
+
+PublicSuffix::List.default.reindex!

 module Twingly
   class URL
```

Had to add `reindex: false` to `add` (and `PublicSuffix::List.default.reindex!`) otherwise it took ~5 seconds to load the file into pry 😄

dentarg:
@roback `<<` is alias for `add`, https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/lib/public_suffix/list.rb#L195

dentarg:
But I like the suggestions above about our own list, looks nice to be able to pass `default_rule: nil` when creating the list

roback:
> @roback << is alias for add, https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/lib/public_suffix/list.rb#L195

Aha, I could have kept the `<<` then :)

lib/twingly/url.rb
dentarg: I used the same expressions as https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/test/psl_test.rb#L37-L38, but that expression might be constructed to handle more variants (https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/test/tests.txt)

we can check if we get the same count using `^`

dentarg:
> looks nice to be able to pass default_rule: nil when creating the list

I misread that, we still pass it when doing `#parse`

dentarg:
> Aha, I could have kept the << then :)

Oops, no, we need to use `#add`, I thought `reindex: false` was the default, but it's `reindex: true`.

dentarg:
> This doesn't look right:

Good catch, I used your fix

dentarg:
Hmm

https://en.wikipedia.org/wiki/Internationalized_domain_name#ToASCII_and_ToUnicode

> The conversions between ASCII and non-ASCII forms of a domain name are accomplished by algorithms called ToASCII and ToUnicode. These algorithms are not applied to the domain name as a whole, but rather to individual labels. For example, if the domain name is www.example.com, then the labels are www, example, and com. ToASCII or ToUnicode are applied to each of these three separately.

I wonder if we do it correctly here...

lib/twingly/url.rb
```diff
@@ -6,7 +6,19 @@
 require_relative "url/error"
 require_relative "version"
 
-PublicSuffix::List.private_domains = false
+PublicSuffix::List.default = PublicSuffix::List.parse(
+  File.read(PublicSuffix::List::DEFAULT_LIST_PATH), private_domains: false)
+
+# Add ASCII form of all internationalized domain names to the public suffix list
+PublicSuffix::List.default.
+  map { |rule| Addressable::IDNA.to_ascii(rule.value) }.
```
walro: We could try with both "pure" and "libidn" implementations just so we know we won't run into something exciting once we drop libidn.

dentarg: I tried manually with

```diff
diff --git a/lib/twingly/url.rb b/lib/twingly/url.rb
index 4193abd..9dd4581 100644
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -1,5 +1,5 @@
 require "addressable/uri"
-require "addressable/idna/native"
+# require "addressable/idna/native"
 require "public_suffix"

 require_relative "url/null_url"
@@ -29,7 +29,7 @@ module Twingly
     ERRORS_TO_EXTEND = [
       Addressable::URI::InvalidURIError,
       PublicSuffix::DomainInvalid,
-      IDN::Idna::IdnaError,
+      # IDN::Idna::IdnaError,
     ]

     private_constant :ACCEPTED_SCHEMES, :ENDS_WITH_SLASH, :ERRORS_TO_EXTEND
diff --git a/twingly-url.gemspec b/twingly-url.gemspec
index f06d0d9..89f14cd 100644
--- a/twingly-url.gemspec
+++ b/twingly-url.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|

   s.add_dependency "addressable", "~> 2"
   s.add_dependency "public_suffix", "~> 2"
-  s.add_dependency "idn-ruby", "~> 0.1"
+  # s.add_dependency "idn-ruby", "~> 0.1"

   s.add_development_dependency "rake", "~> 10"
   s.add_development_dependency "rspec", "~> 3"
```

```
~/twingly/ruby/twingly-url public_suffix-2*
$ bundle show | grep idn-ruby
```

and all specs pass, but I'm not sure it says much, as we just have a few IDNs in our tests cases. But I think it's okay. :)

lib/twingly/url.rb
jage: I think `/\Axn\-\-/i` is the correct regex.

> The ToASCII and ToUnicode operations MUST recognize the ACE prefix in a case-insensitive manner.

> The ACE prefix for IDNA is "xn--" or any capitalization thereof.

https://tools.ietf.org/html/rfc3490#section-5

dentarg:
Continuing the hidden conversation at https://github.com/twingly/twingly-url/pull/90#discussion_r78185929 here

Interesting results, the counts are different


````ruby
[1] pry(main)> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name =~ /xn\-\-/ }.count
=> 814
[2] pry(main)> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name =~ /\Axn\-\-/i }.count
=> 812
[3] pry(main)> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name =~ /^xn\-\-/i }.count
=> 812
````

**Comment by dentarg** [comment]
```ruby
[6] pry(main)> a_814 - a_812
=> ["sande.xn--mre-og-romsdal-qqb.no", "sande.xn--mre-og-romsdal-qqb.no"]
```

dentarg:
Hmm

```ruby
[7] pry(main)> (a_814 - a_812).uniq
=> ["sande.xn--mre-og-romsdal-qqb.no"]
[8] pry(main)> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name =~ /xn\-\-/ }.uniq.count
=> 407
[9] pry(main)> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name =~ /\Axn\-\-/i }.uniq.count
=> 406
```

dentarg:
Oh, my bad, I'm in `bundle console`, so I've already loaded the code once... :)

jage:
> Interesting results, the counts are different

Shouldn't we check each part? (https://github.com/twingly/twingly-url/pull/90#issuecomment-246268668)

```Ruby
> PublicSuffix::List.default.map { |rule| Addressable::IDNA.to_ascii(rule.value) }.select { |name| name.split(".").any? { |part| part =~ /\Axn\-\-/i } }.count
814
```

dentarg:
Yes

jage:
> Shouldn't we check each part? (#90 (comment))

It could be done with a regex, like `/(\A|\.)xn\-\-/i`, but I think I prefer the explicit split.

dentarg:
I'm thinking of using http://www.rubydoc.info/gems/public_suffix/PublicSuffix/Domain#name_to_labels-class_method instead of splitting manually.

dentarg:
In any case, `sande.xn--mre-og-romsdal-qqb.no` is `sande.møre-og-romsdal.no` in [the list](https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/data/list.txt).

dentarg:
> Shouldn't we check each part? (#90 (comment))

Both versions of `.to_ascii` in Addressable does the splitting to individual parts, so I don't think we need to, our should do it (our code would be more complex).

* https://github.com/sporkmonger/addressable/blob/addressable-2.4.0/lib/addressable/idna/pure.rb#L64-L88
* https://github.com/sporkmonger/addressable/blob/addressable-2.4.0/lib/addressable/idna/native.rb

dentarg:
Sorry, I mixed up things. We run `.to_ascii`, and then we look for candidates to extend the PSL with, and when looking for candidates, we need to examine each part when using `\A` in the regexp (which should be more correct). So yes, we should check each part.

dentarg:
I think I'm done here. I suggest we release this as 4.2.1 (bugfix).

dentarg:
Discussed with @jage, will extract the Public Suffix init code to it's own class.

lib/twingly/public_suffix_list.rb
```diff
@@ -0,0 +1,35 @@
...
+        new_rule = PublicSuffix::Rule.factory(punycoded_name)
+        list.add(new_rule, reindex: false)
+      end
+
+      list.reindex!
+
+      list
+    end
+
+    private_class_method def self.punycoded_names(list)
```
jage: An alternative could be:

```Ruby
private_class_method \
def self.punycoded_names(list)
  # ..
end
```

Not sure which I prefer though (probably one with newline).

dentarg: Okay

dentarg: ba6c5a4

dentarg:
@jage WDYT? Merge?

jage:
I think we need to add a comment (in the PR) as to what changes here. Previously we had som consistency as to what `tld` and `domain` returned (always unicode), now they return unicode if the input is unicode and ascii if the input is ascii. To be honest I think I would prefer that we always returned the same format, no matter what the input is.

Previous version:

```Ruby
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3fd876a72dd8 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3fd8767d277c https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)>
[4] pry(main)> u1.to_s
=> "https://www.foo.ایران.ir/bar"
[5] pry(main)> u2.to_s
=> "https://www.foo.xn--mgba3a4f16a.ir/bar"
[6] pry(main)>
[7] pry(main)> u1.domain
=> "foo.ایران.ir"
[8] pry(main)> u2.domain
=> "foo.ایران.ir"
[9] pry(main)>
[10] pry(main)> u1.tld
=> "ایران.ir"
[11] pry(main)> u2.tld
=> "ایران.ir"
```

This version:

```Ruby
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3fc621fd5454 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3fc621fb855c https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)>
[4] pry(main)> u1.to_s
=> "https://www.foo.ایران.ir/bar"
[5] pry(main)> u2.to_s
=> "https://www.foo.xn--mgba3a4f16a.ir/bar"
[6] pry(main)>
[7] pry(main)> u1.domain
=> "foo.ایران.ir"
[8] pry(main)> u2.domain
=> "foo.xn--mgba3a4f16a.ir"
[9] pry(main)>
[10] pry(main)> u1.tld
=> "ایران.ir"
[11] pry(main)> u2.tld
=> "xn--mgba3a4f16a.ir"
```

dentarg:
> To be honest I think I would prefer that we always returned the same format, no matter what the input is.

Isn't that what we have normalization for?

jage:
> Isn't that what we have normalization for?

Yes and no, normalization is useful to compare URLs (to avoid duplicates). But if I want to print the URL in my web UI, shouldn't we print the same string regardless of the input was in unicode or ascii?

dentarg:
> Previously we had som consistency as to what tld and domain returned (always unicode), 

This just happened to be the case, it wasn't documented.

> now they return unicode if the input is unicode and ascii if the input is ascii.

That is true, we can add tests for that.

jage:
> This just happened to be the case, it wasn't documented.

That doesn't change the fact.

dentarg:
> But if I want to print the URL in my web UI, shouldn't we print the same string regardless of the input was in unicode or ascii?

Yes, that makes sense, but fortunately, we don't have the need to do that right now, so I think we are okay. I don't think that should be the default behaviour though, I think it should be a function that the user explicitly uses. Much like https://github.com/twingly/twingly-url/issues/71, "get URL (or parts of) in Unicode".

dentarg:
> Previously we had som consistency as to what tld and domain returned (always unicode), now they return unicode if the input is unicode and ascii if the input is ascii.

One could argue we are more consistent now, what the methods return are based on the input. E.g. `#host` wasn't always unicode before.

Previous version:

```ruby
$ bundle console
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3ff0792d88b4 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3ff079d181f8 https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)> u1.host
=> "www.foo.ایران.ir"
[4] pry(main)> u2.host
=> "www.foo.xn--mgba3a4f16a.ir"
```

This version:

```ruby
$ bundle console
Resolving dependencies...
[1] pry(main)> u1 = Twingly::URL.parse("https://www.foo.ایران.ir/bar")
=> #<Twingly::URL:0x3fe381504f48 https://www.foo.ایران.ir/bar>
[2] pry(main)> u2 = Twingly::URL.parse("https://www.foo.xn--mgba3a4f16a.ir/bar")
=> #<Twingly::URL:0x3fe381186484 https://www.foo.xn--mgba3a4f16a.ir/bar>
[3] pry(main)> u1.host
=> "www.foo.ایران.ir"
[4] pry(main)> u2.host
=> "www.foo.xn--mgba3a4f16a.ir"
```

As said, I can add tests for this.

dentarg:
Just a note about the [Nameprep](https://en.wikipedia.org/wiki/Nameprep) algorithm, from https://en.wikipedia.org/wiki/Internationalized_domain_name#ToASCII_and_ToUnicode

> The function ToUnicode reverses the action of ToASCII, stripping off the ACE prefix and applying the Punycode decode algorithm. It does not reverse the Nameprep processing, since that is merely a normalization and is by nature irreversible.

dentarg:
> I suggest we release this as 4.2.1 (bugfix).

Perhaps not, as @jage [showed](https://github.com/twingly/twingly-url/pull/90#issuecomment-247021829), we will change the behaviour for some methods, so 5.0.0 might make sense.

dentarg:
Could have been part of the commit message in my latest commit (15414ed), but I can write it here.

* Apparently, Addressable's `#origin` is always normalized :-), see https://github.com/sporkmonger/addressable/blob/addressable-2.4.0/lib/addressable/uri.rb#L1224-L1240

* `макдональдс.рф` is `mcdonalds.ru` in Cyrillic :-) I got it from https://uxmag.com/articles/a-url-in-any-language, as I wanted something as real and working as possible (be able to use in the browser), but I didn't find any real URL with IDN subdomain, IDN domain and IDN TLD, so I faked it by adding `räksmörgås` myself

README.md
```diff
@@ -37,6 +37,41 @@ url.host                # => "www.twingly.co.uk"
...
+url.scheme              # => "http"
+url.trd                 # => "räksmörgås"
+url.sld                 # => "макдональдс"
+url.tld                 # => "рф"
+url.ttld                # => "рф"
+url.domain              # => "макдональдс.рф"
+url.host                # => "räksmörgås.макдональдс.рф"
+url.origin              # => "http://xn--rksmrgs-5wao1o.xn--80aalb1aicli8a5i.xn--p1ai"
+url.path                # => "/foo"
+url.without_scheme      # => "//räksmörgås.макдональдс.рф/foo"
```
jage: How about we print the normalized version as well (for all parts).

```
url.without_scheme
url.normalized.without_scheme
```

Since they return different strings.

dentarg: Sure, I was thinking of showing normalization examples of these URLs as another section below, but maybe inline is better?

jage: If it's the same URL we compare non-normalized and normalized, I think inline (alternating) is easier to read, my assumption is that the reader wants to understand the potential difference between a normalized and a non-normalized URL.

dentarg: Would be great to not have to repeat all the parts for the normalized version, so I'm thinking a new section is easier, WDYT?

jage: Yes, with the current structure it might become too many rows. We can add a new section now.

I can open an issue about improving the output, I'm thinking some of the behavior needs a table:

| Method | Non-normalized | Normalized|
|---------|-----------------|------------|
| `.host` | `räksmörgås.макдональдс.рф` | `xn--rksmrgs-5wao1o.xn--80aalb1aicli8a5i.xn--p1ai ` |

WDYT?

walro: Table seems like a nice solution!

dentarg: Hmm, I can add them as you suggested. I think it will make the most clear examples.

dentarg: Heh, GitHub failed to show me your comments, saw them now when I did a full page reload...

jage:
LGTM!

> Perhaps not, as @jage showed, we will change the behaviour for some methods, so 5.0.0 might make sense.

Yes, agree that this is a major version bump.

README.md
dentarg: I think our specs also could benefit from the "table-way" of doing things.

PR #95, opened 2016-09-15
DRY up to urls example

jage:
This will print userinfo, user and password for all URLs, which it didn't do before.


jage:
The README.md should be updated if this is merged.

dentarg:
@jage LGTM, you can go ahead and update the README and then we can merge it

PR #99, opened 2016-09-19
Make sure we always read PSL data as UTF-8

### LINKED ISSUE #98: "ArgumentError: invalid byte sequence in US-ASCII" when parsing the public suffix list
Seen in [an application](https://github.com/twingly/wally/issues/56) not using `LANG=en_US.UTF-8`:

```
ArgumentError: invalid byte sequence in US-ASCII
  from public_suffix/list.rb:89:in `strip!'
  from public_suffix/list.rb:89:in `block (2 levels) in parse'
  from public_suffix/list.rb:88:in `each_line'
  from public_suffix/list.rb:88:in `block in parse'
  from public_suffix/list.rb:128:in `initialize'
  from public_suffix/list.rb:87:in `new'
  from public_suffix/list.rb:87:in `parse'
  from twingly/public_suffix_list.rb:12:in `with_punycoded_names'
  from twingly/url.rb:15:in `<class:URL>'
  from twingly/url.rb:11:in `<module:Twingly>'
  from twingly/url.rb:10:in `<top (required)>'
  from twingly/url/hasher.rb:3:in `require_relative'
  from twingly/url/hasher.rb:3:in `<top (required)>'
  from /home/deploy/twingly-wally/releases/20160916113112/lib/models/...
```

Fortunately, we read the list ourselves, so we can read it with the correct encoding.

(public_suffix just trusts the default encoding: https://github.com/weppos/publicsuffix-ruby/blob/v2.0.2/lib/public_suffix/list.rb#L52)

dentarg:
Close #98.

### DIFF
--- a/lib/twingly/public_suffix_list.rb
+++ b/lib/twingly/public_suffix_list.rb
@@ -1,3 +1,4 @@
+require "addressable/idna"
 require "public_suffix"
 
 module Twingly
@@ -7,8 +8,9 @@ module Twingly
     private_constant :ACE_PREFIX
 
     # Extend the PSL with ASCII form of all internationalized domain names
-    def self.with_punycoded_names
-      list_data = File.read(PublicSuffix::List::DEFAULT_LIST_PATH)
+    def self.with_punycoded_names(encoding: Encoding::UTF_8)
+      list_path = PublicSuffix::List::DEFAULT_LIST_PATH
+      list_data = File.read(list_path, encoding: encoding)
       list = PublicSuffix::List.parse(list_data, private_domains: false)
 
       punycoded_names(list).each do |punycoded_name|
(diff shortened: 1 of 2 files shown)


lib/twingly/public_suffix_list.rb
```diff
@@ -1,3 +1,4 @@
+require "addressable/idna"
```
walro: Was this missing from before? I don't see how it's used now?

dentarg: This was missing from before, and the specs will fail without it (because the spec runs this file directly – it was required indirectly before)

dentarg: We do `Addressable::IDNA.to_ascii` here (from before, not added in this PR – but there were no specs then, so it made sense to include it in the commit creating the spec file)

walro: Thanks, didn't check the full file before so I didn't spot the call to to_ascii.

walro:
LGTM

PR #102, opened 2016-11-03
Drop libidn

### LINKED ISSUE #66: JRuby compatibility (drop libidn requirement)
Since the introduction of the `idn-ruby` gem in #63 we can no longer run under the Java runtime. Some ideas from the top of my head:
- Stop using the gem
- Is it possibly to have conditionals in the Gemfile? As in, don't use `idn-ruby` for Java platform.
- If above is possible it would mean that we'd handle urls different depending upon runtime which is not that desirable.

dentarg:
Instead, we use the pure Ruby IDNA solution in addressable.

Close #66.

### DIFF
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -114,6 +114,16 @@ describe Twingly::URL do
       end
     end
 
+    context "IDNA 2003 vs IDNA 2008" do
+      describe "if you normalize a URL with the domain \"straße.de\"" do
+        let(:test_url) { "http://straße.de" }
+
+        it "the domain of the normalized URL should be \"xn--strae-oqa.de\"" do
+          expect(subject.normalized.domain).to eq("xn--strae-oqa.de")
+        end
+      end
+    end
+
     context "when given badly encoded input" do
       let(:badly_encoded_url) { "http://abc.se/öあ\x81b\xE3" }
       let(:expected)          { "http://abc.se/öあ\uFFFDb\uFFFD" }
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -42,9 +42,6 @@ def invalid_urls
     # https://github.com/sporkmonger/addressable/issues/224
     "http://some_site.net%C2",
     "http://+%D5d.some_site.net",
-
-    # Triggers IDN::Idna::IdnaError: Output would be too large or too small (5)
-    "http://AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallumEst.com",
   ]
 end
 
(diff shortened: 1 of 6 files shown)


spec/lib/twingly/url_spec.rb
```diff
@@ -102,6 +114,16 @@ def valid_urls
       end
     end
 
+    context "IDNA 2003 vs IDNA 2008" do
```
walro: context and its does not really follow suit with rest of the spec file. How about something like:

```ruby
context "when given IDN URL containing ß" do # could generalize but it's early in the morning and brain is slow
 ...
  it "should conform to the IDNA 2008 standard" do
    ...
  end
end
```

Should also be placed under the describe for `#normalized`, see for instance https://github.com/twingly/twingly-url/blob/77631f74eb5f52e42e009537d0963eaea33ac0a8/spec/lib/twingly/url_spec.rb#L388

walro:
Could cleanup Ansible too: https://github.com/twingly/ansible/commit/c8d999f8464fc7a81a51d18d54c690069a7719fc. I don't think we should uninstall though, I am pretty sure other things depend on it 😄

spec/lib/twingly/url_spec.rb
dentarg: I know the wording doesn't follow what we have... it is hard writing these things. In the code, you have "it", but in the spec output, you don't have "it"... so should you take it into consideration or not?

Now I just used `context` as a "title" and then added an sentence that would read well (IMHO) in the spec output:

```
    IDNA 2003 vs IDNA 2008
      if you normalize a URL with the domain "straße.de"
        the domain of the normalized URL should be "xn--strae-oqa.de"
```

dentarg:
Re: ansible, yes, we should clean up, but we can discuss that over there as it's really not that related to the gem itself

walro:
> Re: ansible, yes, we should clean up, but we can discuss that over there as it's really not that related to the gem itself

Of course not, it was just a reminder.

spec/lib/twingly/url_spec.rb
walro: You should add the `it` yourself when reading, it seems to be what RSpec does in their examples, check for instance:

https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/example-groups/basic-structure-describe-it where you have `it "does something"` and https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/example-groups/shared-examples where you can find `it "should return...."`

walro: > considering we want to do things like #71. That I'm using normalized here is just me "cheating"... :)

I don't think it's cheating (as it seems like now, we need to `#normalize`to trigger this, if I am not mistaken) I believe it should be put under that description. Easier to find and it's what we do currently for the most parts. 

I think it's hard to reason about #71 at this stage as it makes everything more complex, let's take that discussion when we get there 😄

walro: > Now I just used context as a "title" and then added an sentence that would read well (IMHO) in the spec output:

What makes sense to me, when it comes to context, is that you are describing a setting. It's not obvious in a library like this, so I agree it's hard too. It would be more obvious if we dealt with higher level stuff like "when logged in" vs "when not logged in".

jage:
> Close #66.

Should we enable Travis CI tests for JRuby then?

dentarg:
> Should we enable Travis CI tests for JRuby then?

Good idea, will do

spec/lib/twingly/url_spec.rb
walro: BTW, once upon a time I read http://betterspecs.org/ to try to figure out how to use contexts, describes and its.

spec/lib/twingly/url_spec.rb
```diff
@@ -63,6 +60,21 @@ def valid_urls
...
+    # when used with Addressable and libidn (IDNA 2003)
+    # but considered valid with the pure Ruby IDNA solution in Addressable.
+    #
+    # This inconsistency has as been reported at
+    # https://github.com/sporkmonger/addressable/issues/239
+    #
+    # Long names like this will probably not resolv, see comments at
+    # https://github.com/twingly/twingly-url/issues/66#issuecomment-246349029
+    # but for now as addressable considers them valid, we do the same.
+    "http://AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallumEst.com",
```
dentarg: > This can jump back up again once Addressable 2.5.0 is released with sporkmonger/addressable@c73810f

Nope, I got that wrong too. With Addressable 2.5.0 and libidn, the URL is now considered valid.

PR #103, opened 2016-11-04
Switch to logical requires

dentarg:
Avdi in Ruby Tapas 242:

> Setting up a gem to use logical require paths internally is even
> easier. Since RubyGems automatically adds the gem’s lib directories to
> the load path, in that case we don’t need to mess around with the load
> path at all.


dentarg:
WIP because I didn't look at the specs

walro:
Are we gonna finish this one up?

dentarg:
I hope so :)

dentarg:
> WIP because I didn't look at the specs

Do we need specs?

walro:
> Do we need specs?

Not sure what the original comment was referring to, but I don't think we need specs to ensure we use require properly. I thought it meant remove `require_relative` from specs.

walro:
> I thought it meant remove require_relative from specs.

We don't seem to have any left.

dentarg:
Yeah, the tests we have should be sufficient for this change

dentarg:
Build failed due to network problems in the Travis CI container (time out when fetching gems), restarted

PR #105, opened 2016-11-05
Remove ruby-prof from dev dependencies

### LINKED ISSUE #92: Unfortunate require in Rakefile (profile task), affect specs
In our [Rakefile](https://github.com/twingly/twingly-url/blob/v4.2.0/Rakefile) we have a task that does `require_relative "lib/twingly/url"`:

``` ruby
namespace :profile do
  require_relative "profile/profile"

  desc "Profile"
  task :normalize do |task|
    require_relative "lib/twingly/url"

    Profile.measure "normalizing a short URL", 1000 do
      Twingly::URL.parse('http://www.duh.se/').normalized
    end
  end
end
```

This require could hide problems that we try to catch in our tests, because even if we wont run that particular task, the require is evaluated when doing anything with `rake`, like running our specs.

dentarg:
Was problematic for a number of reasons, see #92.

Close #92.

### DIFF
--- a/profile/Gemfile
+++ b/profile/Gemfile
@@ -1,0 +1,5 @@
+source "https://rubygems.org/"
+
+gem "rake"
+gem "ruby-prof"
+gem "twingly-url", path: "../"
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,8 @@ rdoc
 spec/reports
 test/tmp
 test/version_tmp
-tmp
+/tmp
+/profile/tmp
 
 # YARD artifacts
 .yardoc
--- a/README.md
+++ b/README.md
@@ -180,7 +180,9 @@ Run tests with
 
 You can get some profiling by running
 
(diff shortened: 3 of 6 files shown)


profile/Gemfile
```diff
@@ -0,0 +1,5 @@
+source "https://rubygems.org/"
+
+gem "rake"
+gem "ruby-prof"
```
walro: Slight difference from before, we'll allow any version of rake and ruby-prof now. Possibly not a bad thing though.

dentarg:
> we'll allow any version of rake and ruby-prof now

I think that is a very good change. We can always lock it down later if we discover the need for it.

profile/Gemfile
dentarg: ruby-prof is still on versions 0.x so in practice no difference from what we had, until there is ruby-prof 1.x :)

PR #106, opened 2016-11-05
Tag exceptions

### LINKED ISSUE #31: Rework exceptions

dentarg:
Make it easy to rescue any error from Twingly::URL.parse with `rescue Twingly::URL::Error`.

Close #31.

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -30,6 +30,9 @@ module Twingly
         internal_parse(potential_url)
       rescue Twingly::URL::Error, Twingly::URL::Error::ParseError => error
         NullURL.new
+      rescue Exception => error
+        error.extend(Twingly::URL::Error)
+        raise
       end
 
       def internal_parse(potential_url)
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -85,6 +85,25 @@ describe Twingly::URL do
 
     it { is_expected.to be_a(Twingly::URL) }
 
+    context "when re-reraising errors" do
+      let(:some_exception) { Exception }
+
+      before do
+        allow(described_class)
+          .to receive(:internal_parse)
+          .and_raise(some_exception)
+      end
+
+      it "always tags the error" do
+        expect { subject }.to raise_error do |error|
+          aggregate_failures do
+            expect(error).to be_instance_of(some_exception)
+            expect(error).to be_kind_of(Twingly::URL::Error)
+          end
+        end
+      end
+    end
+
     context "when given valid urls" do
       valid_urls.each do |valid_url|
         it "does not ruin the url \"#{valid_url}\"" do


dentarg:
Not sure how to write a test for this... :)

But I think it works:

```diff
$ git d
diff --git a/lib/twingly/url.rb b/lib/twingly/url.rb
index d62e70e..af39394 100644
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -27,9 +27,13 @@ module Twingly

     class << self
       def parse(potential_url)
+        0/0
         internal_parse(potential_url)
       rescue Twingly::URL::Error, Twingly::URL::Error::ParseError => error
         NullURL.new
+      rescue Exception => error
+        error.extend(Twingly::URL::Error)
+        raise
       end

       def internal_parse(potential_url)
```

```ruby
$ bundle console
Resolving dependencies...
[1] pry(main)> begin
[1] pry(main)*   Twingly::URL.parse("foo")
[1] pry(main)* rescue Twingly::URL::Error => error
[1] pry(main)*   warn "Error in Twingly::URL"
[1] pry(main)* end
Error in Twingly::URL
=> nil
[2] pry(main)> error
=> #<ZeroDivisionError: divided by 0>
[3] pry(main)> Twingly::URL.parse("foo")
ZeroDivisionError: divided by 0
from /Users/dentarg/twingly/ruby/twingly-url/lib/twingly/url.rb:30:in `/'
[4] pry(main)> wtf??
Exception: ZeroDivisionError: divided by 0
--
 0: /Users/dentarg/twingly/ruby/twingly-url/lib/twingly/url.rb:30:in `/'
 1: /Users/dentarg/twingly/ruby/twingly-url/lib/twingly/url.rb:30:in `parse'
```

walro:
> Not sure how to write a test for this... :)

Maybe stubbing `.internal_parse` to throw some error could work for a test case

walro:
Could work, but it somewhat couples tests with implementation (you need to know that `internal_parse` is called by `parse`). Personally I think the benefit outweighs the drawback.

```ruby
index a45ba57..ffc420d 100644
@@ -93,6 +93,18 @@ describe Twingly::URL do
       end
     end
 
+    context "when re-reraising errors" do
+      before do
+        allow(described_class)
+          .to receive(:internal_parse)
+          .and_raise(ZeroDivisionError)
+      end
+
+      it "always tags the error" do
+         expect { subject }.to raise_error(Twingly::URL::Error)
+      end
+    end
+
     context "when given bad input" do
       invalid_urls.each do |invalid_url|
         it "returns a NullURL for \"#{invalid_url}\"" do
```

PR #107, opened 2016-11-06
Depend on addressable 2.5.0 and public_suffix 2.0.3

dentarg:
From https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md#addressable-250

> - dropping support for Ruby 1.9
> - adding support for Ruby 2.4 preview
> - add support for public suffixes and tld; first runtime dependency
> - hostname escaping should match RFC; underscores in hostnames no
>   longer escaped
> - paths beginning with // and missing an authority are now considered
>   invalid
> - validation now also takes place after setting a path
> - handle backslashes in authority more like a browser for
>   `heuristic_parse`
> - unescaped backslashes in host now raise an `InvalidURIError`
> - `merge!`, `join!`, `omit!` and `normalize!` don't disable deferred
>   validation
> - `heuristic_parse` now trims whitespace before parsing
> - host parts longer than 63 bytes will be ignored and not passed to
>   libidn
> - normalized values always encoded as UTF-8

From https://github.com/weppos/publicsuffix-ruby/blob/master/CHANGELOG.md#release-203

> - CHANGED: Updated definitions.

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -59,7 +59,6 @@ module Twingly
         else
           potential_url = String(potential_url)
           potential_url = potential_url.scrub
-          potential_url = potential_url.strip
 
           Addressable::URI.heuristic_parse(potential_url)
         end
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -61,20 +61,6 @@ def valid_urls
     "http://:@blog.twingly.com/",
     "https://www.foo.ایران.ir/bar",
     "https://www.foo.xn--mgba3a4f16a.ir/bar",
-
-    # The following URL triggers
-    #
-    #   IDN::Idna::IdnaError: Output would be too large or too small (5)
-    #
-    # when used with Addressable and libidn (IDNA 2003)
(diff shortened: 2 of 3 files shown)


dentarg:
Now that the behaviour is the same with addressable with and without libidn, I'm thinking the text I added in #102 is outdated... but I don't know what to write instead. Maybe nothing?

```
    # The following URL triggers
    #
    #   IDN::Idna::IdnaError: Output would be too large or too small (5)
    #
    # when used with Addressable and libidn (IDNA 2003)
    # but considered valid with the pure Ruby IDNA solution in Addressable.
    #
    # This inconsistency has as been reported at
    # https://github.com/sporkmonger/addressable/issues/239
    #
    # Long names like this will probably not resolv, see comments at
    # https://github.com/twingly/twingly-url/issues/66#issuecomment-246349029
    # but for now as addressable considers them valid, we do the same.
    "http://AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallumEst.com",
```

walro:
> but I don't know what to write instead. Maybe nothing?

"Nothing" seems right to me. We can keep the URL around if behavior would change in addressable in the future.

Issue #104, opened 2016-11-05
Dependencies not locked correctly


dentarg:
https://github.com/twingly/twingly-url/commit/b61f962115c7aa6e85f6da9c80dcac102d924e2e got it all wrong. Discovered as Addressable 2.5.0 was released yesterday (all Travis build failed).

```
$ rm Gemfile.lock

~/twingly/ruby/twingly-url master
$ bundle
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 10.5.0
Using coderay 1.1.1
Using diff-lcs 1.2.5
Using idn-ruby 0.1.0
Using method_source 0.8.2
Using slop 3.6.0
Using rspec-support 3.5.0
Using bundler 1.13.3
Using pry 0.10.4
Installing ruby-prof 0.16.2 with native extensions
Installing public_suffix 2.0.3
Installing rspec-core 3.5.4
Using rspec-expectations 3.5.0
Using rspec-mocks 3.5.0
Using rspec 3.5.0
Installing addressable 2.5.0
Using twingly-url 5.0.1 from source at `.`
Bundle complete! 5 Gemfile dependencies, 17 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
```

```
$ rm Gemfile.lock

~/twingly/ruby/twingly-url master*
$ git d
diff --git a/twingly-url.gemspec b/twingly-url.gemspec
index e5b5f6f..0034d19 100644
--- a/twingly-url.gemspec
+++ b/twingly-url.gemspec
@@ -12,8 +12,8 @@ Gem::Specification.new do |s|
   s.license     = "MIT"
   s.required_ruby_version = "~> 2.2"

-  s.add_dependency "addressable", "~> 2.4", ">= 2.4.0"
-  s.add_dependency "public_suffix", "~> 2.0", ">= 2.0.2"
+  s.add_dependency "addressable", "~> 2.4", "<= 2.4.0"
+  s.add_dependency "public_suffix", "~> 2.0", "<= 2.0.2"
   s.add_dependency "idn-ruby", "~> 0.1"

   s.add_development_dependency "rake", "~> 10"

~/twingly/ruby/twingly-url master*
$ bundle
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 10.5.0
Using addressable 2.4.0
Using coderay 1.1.1
[...]

PR #117, opened 2018-02-09
Update PublicSuffix and Addressable

roback:
Unfortunately it doesn't look like this fixes any of our issues, but since it made the profiling run a bit faster (and the fact that the tests didn't break) I made a PR of this anyway.

Profiling "total run time" went from `1.6663s` to `1.4801s`.

Some related links:

* https://github.com/weppos/publicsuffix-ruby/pull/130
* https://github.com/weppos/publicsuffix-ruby/pull/133
* https://github.com/sporkmonger/addressable/pull/267

### DIFF
--- a/lib/twingly/public_suffix_list.rb
+++ b/lib/twingly/public_suffix_list.rb
@@ -15,17 +15,15 @@ module Twingly
 
       punycoded_names(list).each do |punycoded_name|
         new_rule = PublicSuffix::Rule.factory(punycoded_name)
-        list.add(new_rule, reindex: false)
+        list.add(new_rule)
       end
-
-      list.reindex!
 
       list
     end
 
     private_class_method \
     def self.punycoded_names(list)
-      names = list.map { |rule| Addressable::IDNA.to_ascii(rule.value) }
+      names = list.each.map { |rule| Addressable::IDNA.to_ascii(rule.value) }
       names.select { |name| punycoded_name?(name) }
     end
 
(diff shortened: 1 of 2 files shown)


walro:
We could take the opportunity and think about: https://github.com/twingly/twingly-url/issues/104#issuecomment-361502143 

My experience that going from loose requirements to strict ones made it a bit cumbersome for Bundler to figure out how to update. I would usually end up with `twingly-url` not being updated at all (unless I danced a bit). Not sure what I proposed, allowing patch versions, would be a silver bullet but it would likely help in some cases at least.

roback:
Just a note: We need at least Addressable `2.5.2`, since that is the only version is compatible with public suffix `3.x`: https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md#addressable-252

walro:
We could allow for public_suffix > 2, < 4 and something similar for addressable and let bundler figure out the rest. I am not really sure about how we properly test all combinations though.

roback:
> I am not really sure about how we properly test all combinations though.

Looks like Travis already thought of that: https://docs.travis-ci.com/user/languages/ruby/#Testing-against-multiple-versions-of-dependencies, though we have to create the gemfiles manually.

dentarg:
Let's get this merged and released and then loosen up the dependencies and test with multiple gemfiles.

dentarg:
@texpert done! see https://rubygems.org/gems/twingly-url/versions/5.1.1

texpert:
Awesome, @dentarg , thanks a lot!!

PR #124, opened 2018-05-08
Remove Twingly::URL::Hasher.pingloggerdb_hash

walro:
The database is going away so we won't have any need for this.


walro:
[Sprint52: GDPR – Skrota pinglogger](https://trello.com/c/OKVuTRqV/2397-sprint52-gdpr-skrota-pinglogger)

jage:
Oh noes, breaking change.

walro:
> Oh noes, breaking change.

Doesn't like we use it internally at least. 😄

Issue #123, opened 2018-04-16
Array#uniq does not remove equal Twingly::URLs


walro:
```ruby
[2] pry(main)> url = "https://www.twingly.com"
=> "https://www.twingly.com"
[3] pry(main)> Twingly::URL.parse(url) == Twingly::URL.parse(url)
=> true
[4] pry(main)> [Twingly::URL.parse(url), Twingly::URL.parse(url)].uniq.size
=> 2
```

In the sample above I would expect to get 1 element back.

dentarg:
Great catch, not sure how we can solve it (seems tricky)

I will dump the workaround you used here:

```ruby
$ irb -rtwingly/url
irb(main):001:0> a = Twingly::URL.parse("http://google.com")
=> #<Twingly::URL:0x3fc7cf08b444 http://google.com>
irb(main):002:0> b = Twingly::URL.parse("http://google.com")
=> #<Twingly::URL:0x3fc7cec9e890 http://google.com>
irb(main):003:0> [a, b].uniq(&:to_s)
=> [#<Twingly::URL:0x3fc7cf08b444 http://google.com>]
```

jage:
I think this is how to solve it.

```Diff
diff --git a/lib/twingly/url.rb b/lib/twingly/url.rb
index 20b4437..4999fd6 100644
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "digest"
+
 require "addressable/idna/pure"
 require "addressable/uri"
 require "public_suffix"
@@ -23,6 +25,11 @@ module Twingly
       PublicSuffix::DomainInvalid,
     ].freeze
 
+    # Instantiate digest classes in a thread-safe manner
+    # This is important since we don't know how people will
+    # use this gem (if they require it in a thread safe way)
+    SHA256_DIGEST = Digest(:SHA256)
+
     private_constant :ACCEPTED_SCHEMES
     private_constant :CUSTOM_PSL
     private_constant :STARTS_WITH_WWW
@@ -193,6 +200,14 @@ module Twingly
       self.to_s <=> other.to_s
     end
 
+    def eql?(other)
+      self.hash == other.hash
+    end
+
+    def hash
+      SHA256_DIGEST.digest(self.to_s).unpack1("q")
+    end
+
     def to_s
       addressable_uri.to_s
     end
[...]

PR #126, opened 2018-11-21
Strip URLs of leading and trailing non-breaking space (and space, but we already did)

dentarg:
https://en.wikipedia.org/wiki/Non-breaking_space

Background: Got a big list of URLs (Excel file) and many values in the "URL" column started with the `NO-BREAK SPACE` character.


walro:
Is this on parity with what we do for the .NET counterpart? Because this feels quite specific to a single use-case. Not too sure if the responsibility of this cleansing should be the responsibility of this gem.

 I would be onboard if we decided that the library should strip all leading, and possibly, trailing whitespace characters, but as I lead with, this feels very specific.

dentarg:
> Is this on parity with what we do for the .NET counterpart?

Looks like that:

```
$ ruby -e 'puts "\u00A0https://example.com/foo/\u00A0"' | mono trunk/Deploy/Normalizer/Normalizer.exe
 https://example.com/foo/ ,https://www.example.com/foo,5591070627727872931
```

```
$ ruby -e 'puts "\u0020https://example.com/foo/\u0020"' | mono trunk/Deploy/Normalizer/Normalizer.exe
" https://example.com/foo/ ",https://www.example.com/foo,5591070627727872931
```

> Because this feels quite specific to a single use-case. Not too sure if the cleansing should be the responsibility of this gem. I would be onboard if we decided that the library should strip all leading, and possibly, trailing whitespace characters, but as I lead with, this feels very specific.

I agree that it is specific. After I did this, I realised that if we do this, we probably want to remove all whitespace, but I haven't had the time to try to do it.

Re: trailing whitespace, I saw after I did this that Addressable keeps the non-breaking space:

```
$ ruby -raddressable/uri -e 'p Addressable::URI.parse("https://example.com/foo/\u00A0").to_s'
"https://example.com/foo/ "

$ ruby -raddressable/uri -e 'p Addressable::URI.heuristic_parse("https://example.com/foo/\u00A0").to_s'
"https://example.com/foo/ "
```

`Addressable::URI#normalize` also keeps it:

```
$ ruby -raddressable/uri -e 'p Addressable::URI.heuristic_parse("https://example.com/foo/\u00A0").normalize.to_s'
"https://example.com/foo/%20"
```

The behaviour isn't the same for ordinary space (and `heuristic_parse` and `parse` differs):

```
$ ruby -raddressable/uri -e 'p Addressable::URI.parse("https://example.com/foo/\u0020").to_s'
"https://example.com/foo/ "

$ ruby -raddressable/uri -e 'p Addressable::URI.parse("https://example.com/foo/\u0020").normalize.to_s'
"https://example.com/foo/"

$ ruby -raddressable/uri -e 'p Addressable::URI.heuristic_parse("https://example.com/foo/\u0020").to_s'
"https://example.com/foo/"
```

`Addressable::URI#parse` blows up on any leading space:

```
$ ruby -raddressable/uri -e 'p Addressable::URI.parse("\u00A0https://example.com/foo/").to_s'
/Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:874:in `scheme=': Invalid scheme format:  https (Addressable::URI::InvalidURIError)
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:799:in `block in initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:2355:in `defer_validation'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:796:in `initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `new'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `parse'
  from -e:1:in `<main>'

$ ruby -raddressable/uri -e 'p Addressable::URI.parse("\u0020https://example.com/foo/").to_s'
/Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:874:in `scheme=': Invalid scheme format:  https (Addressable::URI::InvalidURIError)
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:799:in `block in initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:2355:in `defer_validation'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:796:in `initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `new'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `parse'
  from -e:1:in `<main>'
```

`Addressable::URI#heuristic_parse` do handle the ordinary space but not the non-breaking space:

```
$ ruby -raddressable/uri -e 'p Addressable::URI.heuristic_parse("\u0020https://example.com/foo/").to_s'
"https://example.com/foo/"

$ ruby -raddressable/uri -e 'p Addressable::URI.heuristic_parse("\u00A0https://example.com/foo/").to_s'
/Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:874:in `scheme=': Invalid scheme format:  https (Addressable::URI::InvalidURIError)
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:799:in `block in initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:2355:in `defer_validation'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:796:in `initialize'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `new'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:136:in `parse'
  from /Users/dentarg/.gem/ruby/2.4.2/gems/addressable-2.5.2/lib/addressable/uri.rb:207:in `heuristic_parse'
  from -e:1:in `<main>'

```

dentarg:
> Not too sure if the cleansing should be the responsibility of this gem.

I think it should, at least leading whitespace. An alternative could be to put it in `Twingly::URL::Utilities.extract_valid_urls`, but I think `.parse` is nicer, we use it more.

dentarg:
Our current behaviour 

```
$ ruby -rtwingly/url -e 'p Twingly::URL.parse("\u0020https://example.com/foo/").to_s'
"https://example.com/foo/"

$ ruby -rtwingly/url -e 'p Twingly::URL.parse("\u00A0https://example.com/foo/").to_s'
""
```
$ ruby -rtwingly/url -e 'p Twingly::URL::VERSION'
"5.1.1"
```

**Comment by walro** [comment]
```
$ ruby -rtwingly/url -e 'p Twingly::URL.parse("\u00A0https://example.com/foo/").to_s'
""
```

OK, this is pretty bad behavior which I guess got you started.

> I agree that it is specific. After I did this, I realised that if we do this, we probably want to remove all whitespace, but I haven't had the time to try to do it.

It sounds like we agree :)

> I think it should, at least leading whitespace. An alternative could be to put it in Twingly::URL::Utilities.extract_valid_urls, but I think .parse is nicer, we use it more.

When I meant if it should be the responsibility of the gem, I was talking about the specific nbsp case, as pre above - I think it's fine (or even a good thing) that we do it in general

**Comment by dentarg** [comment]
I think it might take more time to make it an option than to understand how breaking it would be. I think we should make it more general as @walro and I have discussed, and if need be, bump major the major version when we release it.

**Comment by dentarg** [comment]
We already do some cleaning so also I don't think it is a good idea to add `clean:`. I prefer to keep the usage simple.

**Comment by walro** [comment]
> I think it might take more time to make it an option than to understand how breaking

Wouldn't a naïve solution work:

```ruby
...
... = clean_input(input) if clean
...
```

walro:
> We already do some cleaning so also I don't think it is a good idea to add `clean:`. I prefer to keep the usage simple.

Ah I see.

jage:
Sure, just to clarify I'm not worried about bumping the major, I'm worried about potential new issues in other systems that might pop up. I.e. and upgrade to twingly-url will be blocked by some other system not liking this. Or that we are "forced" to implement this in more places. But if we look into it, of course this is a good change 👍 

Adding this to `extract_valid_urls` would be much more straight forward IMHO, "just" split or more than space. i.e.. we don't need to worry as much about compatibility (which I agree, might not be an issue here, but I don't _know_). But it's not the same feature, then you'd need to use the utilities method and not parse directly, but just wanted to note this anyway.

dentarg:
Seeing that our .NET library do handle this (https://github.com/twingly/twingly-url/pull/126#issuecomment-442633861) I don't think we should be afraid of implementing it here and start using it.

> I'm worried about potential new issues in other systems that might pop up. I.e. and upgrade to twingly-url will be blocked by some other system not liking this.

I'm not that worried, and as long as we don't [do like this](https://twitter.com/JeanAbayo/status/1089983421408047104), I think we will be fine with our usual approach. We deploy the new version to one system at a time, and observe the outcome. I think that is less work than an extending the API. I'm afraid that it wont be used at all then.

jage:
I think I got stuck on the phrase "Looks like that:". I mean, if it _is_ on par, then it should be fine.

jage:
I'm thinking we should try get this in next week, this should, together with #128 (and maybe #123 ), make for a good release (patch or minor).

jage:
Not sure this should stop this, but just wanted to note a subtle difference:

```Ruby
[1] pry(main)> Twingly::URL.parse(" \u00A0 https://example.com/foo/ \u00A0 ").to_s
=> ""
```

vs.

```Shell
$ ruby -e 'puts " \u00A0 https://example.com/foo/ \u00A0 "' | mono trunk/Deploy/Normalizer/Normalizer.exe
"   https://example.com/foo/   ",https://www.example.com/foo,5591070627727872931
```

dentarg:
Hmm

```
$ ruby -e 'puts " \u00A0 https://example.com/foo/\u{1F4A9}"' | mono trunk/Deploy/Normalizer/Normalizer.exe
"   https://example.com/foo/💩","   https://example.com/foo/💩",9474829912736228245

$ ruby -e 'puts " \u00A0 https://example.com/foo/ \u{1F4A9}" ' | mono trunk/Deploy/Normalizer/Normalizer.exe
"   https://example.com/foo/ 💩","   https://example.com/foo/ 💩",1824702514565726115

$ ruby -e 'puts " \u{1F4A9} https://example.com/foo/ \u{1F4A9}" ' | mono trunk/Deploy/Normalizer/Normalizer.exe
" 💩 https://example.com/foo/ 💩"," 💩 https://example.com/foo/ 💩",6741913218986405642
```

dentarg:
I don't think this PR is in a state to be merged right now.

For starters, as discussed above, it should be more general, and handle more whitespace (maybe we only need to add the regular space to be happy, that should fix the discrepancy seen in https://github.com/twingly/twingly-url/pull/126#issuecomment-460573800).

Further more, I'm not sure that `.parse` should trim trailing whitespace. Maybe the character is part of the URL and it breaks?

https://github.com/twingly/twingly-url/blob/991dbaabba1824a1d4e40c6a002560f858a04c84/lib/twingly/url.rb#L78

When we normalize, we can trim trailing whitespace, to achieve same output as the .NET normalizer.

dentarg:
> I'm not sure that `.parse` should trim trailing whitespace

Ah, we already do trim the regular space.

jage:
> For starters, as discussed above, it should be more general, and handle more whitespace (maybe we only need to add the regular space to be happy, that should fix the discrepancy seen in [#126 (comment)](https://github.com/twingly/twingly-url/pull/126#issuecomment-460573800)).

That can be added later, no need to rewrite everything in one PR. I don't think that's an argument for stopping this, we can just continue the work in a new PR?

> ... we only need to add the regular space to be happy ...

Not sure what you mean? Regulare space is already removed when parsing?

> Further more, I'm not sure that `.parse` should trim trailing whitespace. Maybe the character is part of the URL and it breaks if we remove it?

Yeah, that is a risk.

---

I think we should close this PR if we aren't sure what we want to achieve here, might be better to work out a plan for the `.parse` function in an issue and then focus on the code development? It sounds like some ideas mentioned in this PR shouldn't be in `.parse` at all, but rather in a broader "extract URL(s) from this string".

I've actively put effort validating stuff here today, I thought you wanted this merged. I think this is ok, there is no contract between our implementations and we're more strict here AFAIK than in our .NET code, so the risk of adding this should be small (but not zero). BUT, if we're not sure about this I'd rather see this closed.

dentarg:
> Ah, we already do trim the regular space.

I had a hard time seeing where this was done, but it is the heuristic parsing in Addressable that does it:

```
$ ruby -rbundler/setup -r./lib/twingly/url -e 'p Addressable::URI.heuristic_parse(" https://example.com/foo/ ").to_s'
"https://example.com/foo/"
```

lib/twingly/url.rb
```diff
@@ -62,13 +75,22 @@ def internal_parse(potential_url)
...
+      end
+
+      def strip_whitespace(input)
+        return input unless input.encoding == Encoding::UTF_8
+
+        input.gsub(LEADING_AND_TRAILING_WHITESPACE, "")
+      end
+
       def to_addressable_uri(potential_url)
         if potential_url.is_a?(Addressable::URI)
```
walro: this line can never be anything else than `false`, so we can just remove it. I assume it was for performance once upon a time (do not parse something that's already parsed), but now we want the URIs to be clean so we cannot use a prior parse anyway.

walro: > I assume it was for performance once upon a time

Yes: 219cf37a3d53dfcd7a1c70e7560c508940eec4d6 :)

walro:
(maybe PR title needs a slight tweak btw)

PR #127, opened 2019-01-30
Support: Ruby 2.6, 2.5, 2.4, drop 2.2, 2.3

jage:
2.2 and 2.3 has lost their support.


walro:
Feels a bit weird to drop support for the JRuby series (9.1.x) we're using in some projects. I guess the middle ground would be to just drop 2.2/9.0 support for now.

Concerning JRuby, it seems like they are still offering 9.1 without stuffing it into earlier releases at least: https://www.jruby.org/download

walro:
Moreover, 2.3 loses its support on [2019-03-31](https://www.ruby-lang.org/en/downloads/branches/).

dentarg:
> Feels a bit weird to drop support for the JRuby series (9.1.x) we're using in some project(s).

Yeah, I had that feeling too. I did check, and we are not using twingly-url in our JRuby 9.1.x project (bruce). I think it is unlikely we will start doing it soon.

> Moreover, 2.3 loses its support on [2019-03-31](https://www.ruby-lang.org/en/downloads/branches/).

Yeah, I saw it too (after my commits), that 2.3 is still supported... but as we don't use 2.3 anywhere, I'm ok with dropping it now.

jage:
> Yeah, I saw it too (after my commits), that 2.3 is still supported... but as we don't use 2.3 anywhere, I'm ok with dropping it now.

Yeah, and AFAIK 2.3 is in "security maintenance" (vs. normal maintenance for 2.4).

> Currently in security maintenance phase. We will never backport any bug fixes to 2.3 except security fixes. If a critical security issue is found, we will release an urgent fix for it. We are planning to end the support of the Ruby 2.3 series at the end of March 2019.

https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

PR #128, opened 2019-01-31
Freeze string literals and mutable constants

jage:
I did this with rubocop:

    rubocop --only Style/FrozenStringLiteralComment,Style/MutableConstant --auto-correct

Opted to not add rubocop at this stage, wanted to limit the amount of
work here.

### DIFF
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "spec_helper"
 
 require "twingly/url"
@@ -134,7 +136,7 @@ describe Twingly::URL do
     end
 
     context "when given ASCII input" do
-      let(:ascii_url) { "http://www.twingly.com/öあ".force_encoding("ASCII-8BIT") }
+      let(:ascii_url) { "http://www.twingly.com/öあ".dup.force_encoding("ASCII-8BIT") }
       let(:expected)  { "http://www.twingly.com/öあ" }
       let(:actual)    { described_class.parse(ascii_url).to_s }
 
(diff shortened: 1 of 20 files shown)


spec/lib/twingly/url_spec.rb
```diff
@@ -134,7 +136,7 @@ def valid_urls
     end
 
     context "when given ASCII input" do
-      let(:ascii_url) { "http://www.twingly.com/öあ".force_encoding("ASCII-8BIT") }
+      let(:ascii_url) { "http://www.twingly.com/öあ".dup.force_encoding("ASCII-8BIT") }
```
jage: This broke without the `.dup`, I think this is ok but feel free to suggest something if you know other ways to create the ascii-string.

walro:
Performance benchmarks? 😉

jage:
> Performance benchmarks? 😉

I'm guessing it's hardly noticeable compared to the Addressable/Public-suffix stuff going on under the hood, but if it would be cool to have two more rake tasks, one to measure speed and another to meassure allocated objects. Currently I think we're only profiling method calls IIRC.

jage:
I think this should be extracted to a frozen constant as well, if I understand this correctly there will be a new Regex object on each `normalize_blogspot`-call currently.

https://github.com/twingly/twingly-url/blob/401cbaa49004fff882d4306ab1a84ba787e79cda/lib/twingly/url.rb#L208

walro:
> I'm guessing it's hardly noticeable compared to the Addressable/Public-suffix stuff going on under the hood, but if it would be cool to have two more rake tasks, one to measure speed and another to measure allocated objects. Currently I think we're only profiling method calls IIRC.

I wasn't that serious, but yeah good ideas! :)

PR #129, opened 2019-02-01
Implement uniqueness (hash equality)

### LINKED ISSUE #123: Array#uniq does not remove equal Twingly::URLs

jage:
Will reduce duplicates when using `.uniq`, to get even better result the user could first normalize, then uniq.

Close #123

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "digest"
-
 require "addressable/idna/pure"
 require "addressable/uri"
 require "public_suffix"
@@ -24,11 +22,6 @@ module Twingly
       Addressable::URI::InvalidURIError,
       PublicSuffix::DomainInvalid,
     ].freeze
-
-    # Instantiate digest classes in a thread-safe manner
-    # This is important since we don't know how people will
-    # use this gem (if they require it in a thread safe way)
-    SHA256_DIGEST = Digest(:SHA256)
 
     private_constant :ACCEPTED_SCHEMES
     private_constant :CUSTOM_PSL
@@ -205,7 +198,7 @@ module Twingly
     end
 
     def hash
-      SHA256_DIGEST.digest(self.to_s).unpack1("q")
+      self.to_s.hash
     end
 
     def to_s
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "digest"
+
 require "addressable/idna/pure"
 require "addressable/uri"
 require "public_suffix"
@@ -22,6 +24,11 @@ module Twingly
       Addressable::URI::InvalidURIError,
       PublicSuffix::DomainInvalid,
     ].freeze
+
+    # Instantiate digest classes in a thread-safe manner
+    # This is important since we don't know how people will
+    # use this gem (if they require it in a thread safe way)
+    SHA256_DIGEST = Digest(:SHA256)
 
     private_constant :ACCEPTED_SCHEMES
     private_constant :CUSTOM_PSL
@@ -191,6 +198,14 @@ module Twingly
 
     def <=>(other)
       self.to_s <=> other.to_s
+    end
+
+    def eql?(other)
+      self.hash == other.hash
+    end
+
+    def hash
+      SHA256_DIGEST.digest(self.to_s).unpack1("q")
     end
 
     def to_s
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -194,6 +194,8 @@ module Twingly
     end
 
     def eql?(other)
+      return false unless other.is_a?(self.class)
+
       self.hash == other.hash
     end
 
--- a/lib/twingly/url/null_url.rb
+++ b/lib/twingly/url/null_url.rb
@@ -25,6 +25,8 @@ module Twingly
       end
 
       def eql?(other)
+        return false unless other.is_a?(self.class)
+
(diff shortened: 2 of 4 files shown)


lib/twingly/url.rb
```diff
@@ -193,6 +200,14 @@ def <=>(other)
       self.to_s <=> other.to_s
     end
 
+    def eql?(other)
+      self.hash == other.hash
+    end
+
+    def hash
+      SHA256_DIGEST.digest(self.to_s).unpack1("q")
```
walro: This looks inspired by something, if so, could you share what?

dentarg: The autoping hash? :)

https://github.com/twingly/twingly-url/blob/65c09ae651d6752cd0f53e2def9201b259eb86d1/lib/twingly/url/hasher.rb#L30-L32

walro: Oh, yeah maybe :)

dentarg:
Link to the components in URI: https://github.com/ruby/ruby/blob/e6d1c72bec5c6544e9ae82501a6cdd7460222096/lib/uri/generic.rb#L44-L53

lib/twingly/url.rb
jage: Yeah, exatly. But I didn’t want to use our hasher, since it’s specific to our databases. 

I read the docs for eql? and hash, wich said there should be a deterministic number. Can’t find the docs from my phone right now though. But I basically just implemented it like we usually calculate hashes. 😀

jage: I guess we could use the string hash directly.

jage:
> Couldn't help myself from checking how Ruby does it for the URI class. If I get it correctly they stuff all "components" into an array and piggyback on that array's hash. Feels kind of neat as we'd avoid the usage of digest.

👍🏻

I’m thinking we don’t want “implicit cleaning” by using the components, thus. maybe ‘.to_s.hash’ could be better in this case? Would remove the digest as well.

walro:
> I’m thinking we don’t want “implicit cleaning” by using the components, thus. maybe ‘.to_s.hash’ could be better in this case? Would remove the digest as well.

SGTM!

lib/twingly/url.rb
```diff
@@ -193,6 +200,14 @@ def <=>(other)
       self.to_s <=> other.to_s
     end
 
+    def eql?(other)
+      self.hash == other.hash
```
walro: should we check that `other` is if same type too?

jage: Yeah, or maybe we can delegate it somehow, hmm.  But yeah that sounds reasonable.

jage: Implemented in https://github.com/twingly/twingly-url/pull/129/commits/d30bbd8324ad6bbd5622e88268da87f5714aeb55

lib/twingly/url.rb
jage: New implementation in 3ecf26fc5d626d8ce53ab20a1b8b7366dfab09ec

lib/twingly/url/null_url.rb
```diff
@@ -24,6 +24,16 @@ def <=>(other)
         self.to_s <=> other.to_s
       end
 
+      def eql?(other)
+        return false unless other.is_a?(self.class)
```
walro: I wonder if it's possible to construct a spec that tests this. I tried commenting out the line, all specs still pass.

walro: (In an another way than introducing another test class that implements `hash` in the same way)

jage: Maybe we can do specific `eql?` testing?

```Diff
diff --git a/spec/lib/twingly/url/null_url_spec.rb b/spec/lib/twingly/url/null_url_spec.rb
index 09b9e04..6073698 100644
--- a/spec/lib/twingly/url/null_url_spec.rb
+++ b/spec/lib/twingly/url/null_url_spec.rb
@@ -112,6 +112,11 @@ describe Twingly::URL::NullURL do
       it "should be two unique objects" do
         expect([a,b].uniq).to eq([a,b])
       end
+
+      context ".eql?" do
+        subject { a.eql?(b) }
+        it { is_expected.to eq(false) }
+      end
     end
   end
 end
index e8bb15c..f2f565f 100644
@@ -884,6 +884,11 @@ describe Twingly::URL do
       it "should be two unique objects" do
         expect([a,b].uniq).to eq([a,b])
       end
+
+      context ".eql?" do
+        subject { a.eql?(b) }
+        it { is_expected.to eq(false) }
+      end
     end
   end

```

walro: good idea! switch the `context` to a `describe` though 😄

jage:
So, should we handle this as a strict bugfix, i.e. patch release? I'm thinking this should be released as a patch, the reasoning is that if you write code with `[url1, url2].uniq` you expect the unique results which didn't work before. If you don't run `#uniq` nothing will change.

walro:
Sounds reasonable, not sure what other unreleased stuff we have though, so it might be a minor release anyway (or even major) :)

jage:
> Sounds reasonable, not sure what other unreleased stuff we have though, so it might be a minor release anyway (or even major) :)

Hehe, yup. :)

PR #133, opened 2019-02-06
Accept that parsing downcases the scheme part, allow Addressable 2.6

### LINKED ISSUE #132: Update addressable to 2.6
We're limiting to 2.5 currently:

https://github.com/twingly/twingly-url/blob/c2218e98ff0892e63adc483f4cf1c06e4f01f3ae/twingly-url.gemspec#L17

Did a quick test, but we need to either change our specs or look into why the protocol is updated:

```
Failures:

  1) Twingly::URL.parse when given valid urls does not ruin the url "hTTP://blog.twingly.com/"
     Failure/Error: expect(described_class.parse(valid_url).to_s).to eq(valid_url)

       expected: "hTTP://blog.twingly.com/"
            got: "http://blog.twingly.com/"

       (compared using ==)
     # ./spec/lib/twingly/url_spec.rb:124:in `block (5 levels) in <top (required)>'
```

dentarg:
This is the situation since Addressable 2.6: https://github.com/sporkmonger/addressable/pull/281

Close #132

### DIFF
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -55,7 +55,6 @@ def valid_urls
   [
     "http://blog.twingly.com/",
     "http://blOg.tWingly.coM/",
-    "hTTP://blog.twingly.com/",
     "https://blog.twingly.com",
     "http://3.bp.blogspot.com/_lRbEHeizXlQ/Sf4RdEqCqhI/AAAAAAAAAAw/Pl8nGPsyhXc/s1600-h/images[4].jpg",
     "http://xn--zckp1cyg1.sblo.jp/",
@@ -133,6 +132,13 @@ describe Twingly::URL do
           expect(actual).to be_a(Twingly::URL::NullURL)
         end
       end
+    end
+
+    context "downcases the protocol" do
+      let(:test_url) { "HTTPS://www.twingly.com/" }
+      let(:expected) { "https://www.twingly.com/" }
+
+      it { is_expected.to eq(expected) }
     end
 
     context "when given badly encoded input" do
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -532,7 +532,7 @@ describe Twingly::URL do
       it { is_expected.to eq(expected) }
     end
 
-    context "is able to normalize a url without protocol" do
+    context "is able to normalize a url without the scheme part" do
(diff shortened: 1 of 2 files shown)


dentarg:
Does this require another major bump? 😬 

Something changed, but I don't think it is particular breaking.

spec/lib/twingly/url_spec.rb
```diff
@@ -135,6 +134,13 @@ def leading_and_trailing_whitespace
       end
     end
 
+    context "downcases the protocol" do
```
walro: "downcases the protocol" sounds more like an "it" than an "context". I'd go with something like:

```ruby
context "when given an URL with uppercase protocol" do
  it "downcases the protocol" do
    ...
```

dentarg: I just copied an existing spec! 🙈

dentarg: I agree :-) I also think we should use "scheme" instead of "protocol", I can fix both tomorrow

dentarg: Hmm, after I've looked more closely I think it is a bigger task to use the style you suggest, because there's soo many specs that should be changed. I think if we start mixing, it might make the situation worse.

walro: I think the style I suggested is very prominent in the file? Are we thinking about the same thing?

walro: For example: https://github.com/twingly/twingly-url/blob/c2218e98ff0892e63adc483f4cf1c06e4f01f3ae/spec/lib/twingly/url_spec.rb#L102-L156 is in that style

I think the other style has snuck in lately

dentarg: I gave it a go anyway, but only for the new spec

walro: Ah, I see your change now.

walro:
> Something changed, but I don't think it is particular breaking.

I don't know honestly. But seeing we're **probably** the only ones on version 6 I think we can go with 6.1 or even 6.0.1 here if we decide the previous behavior was a bug :)

dentarg:
We have always downcases the scheme when normalizing, and I would argue that is in that situation it matters, when you want to compare URLs. Not sure what more one can do with URLs, other than "use" them, e.g. connect to them, and that should still work. That's my reasoning at least. :)

walro:
Sounds reasonable.

PR #138, opened 2019-08-27
Handle some common representations of newlines

dentarg:
There are a few more, https://en.wikipedia.org/wiki/Newline#Unicode, maybe we can add them later if need be.

What triggered this was a need to parse a list of URLs where some ended with non-breaking space followed by line-feed:

```ruby
$ irb -rtwingly/url
irb(main):001:0> Twingly::URL.parse("https://www.example.com\u00A0\n")
=> #<Twingly::URL::NullURL:0x00007f8442004100>
irb(main):002:0> Twingly::URL::VERSION
=> "5.1.0"
```

### DIFF
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -70,6 +70,8 @@ def valid_urls
 end
 
 def leading_and_trailing_whitespace
+  line_feed          = "\u000A"
+  carriage_return    = "\u000D"
   non_breaking_space = "\u00A0"
   space              = "\u0020"
 
@@ -79,6 +81,19 @@ def leading_and_trailing_whitespace
     "non-breaking space, space, non-breaking space" => [non_breaking_space, space, non_breaking_space].join,
     "space and non-breaking space"                  => [space, non_breaking_space].join,
     "space, non-breaking space and space"           => [space, non_breaking_space, space].join,
+
+    "non-breaking space and line-feed"              => [non_breaking_space, line_feed].join,
+    "line-feed and non-breaking space"              => [line_feed, non_breaking_space].join,
+    "space and line-feed"                           => [space, line_feed].join,
+    "line-feed and space"                           => [line_feed, space].join,
+
+    "non-breaking space and carriage-return"        => [non_breaking_space, carriage_return].join,
+    "carriage-return and non-breaking space"        => [carriage_return, non_breaking_space].join,
+    "space and carriage-return"                     => [space, carriage_return].join,
+    "carriage-return and space"                     => [carriage_return, space].join,
+
+    "carriage-return and line-feed"                 => [carriage_return, line_feed].join,
+    "line-feed and carriage-return"                 => [line_feed, carriage_return].join,
   }
 end
 
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -70,12 +70,15 @@ def valid_urls
 end
 
 def leading_and_trailing_whitespace
+  non_breaking_space = "\u00A0"
+  space              = "\u0020"
+
   {
-    "non-breaking space and space" => "\u00A0\u0020",
-    "non-breaking space" => "\u00A0",
-    "non-breaking space, space, non-breaking space" => "\u00A0\u0020\u00A0",
-    "space and non-breaking space" => "\u0020\u00A0",
-    "space, non-breaking space and space" => "\u0020\u00A0\u0020",
+    "non-breaking space and space"                  => [non_breaking_space, space].join,
+    "non-breaking space"                            => [non_breaking_space].join,
+    "non-breaking space, space, non-breaking space" => [non_breaking_space, space, non_breaking_space].join,
+    "space and non-breaking space"                  => [space, non_breaking_space].join,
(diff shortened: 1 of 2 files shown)


dentarg:
Added many tests cases, not all fail with the current release version, these are the failures without change in implementation:

```ruby
Failures:

  1) Twingly::URL.parse with url containing leading and trailing: carriage-return and non-breaking space should eq "https://www.example.com/"
     Failure/Error: it { is_expected.to eq(expected) }

       expected: "https://www.example.com/"
            got: #<Twingly::URL::NullURL:0x00007fcdd0261c18>

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"https://www.example.com/"
       +#<Twingly::URL::NullURL:0x00007fcdd0261c18>

     # ./spec/lib/twingly/url_spec.rb:210:in `block (5 levels) in <top (required)>'

  2) Twingly::URL.parse with url containing leading and trailing: line-feed and non-breaking space should eq "https://www.example.com/"
     Failure/Error: it { is_expected.to eq(expected) }

       expected: "https://www.example.com/"
            got: #<Twingly::URL::NullURL:0x00007fcde040e858>

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"https://www.example.com/"
       +#<Twingly::URL::NullURL:0x00007fcde040e858>

     # ./spec/lib/twingly/url_spec.rb:210:in `block (5 levels) in <top (required)>'

  3) Twingly::URL.parse with url containing leading and trailing: non-breaking space and line-feed should eq "https://www.example.com/"
     Failure/Error: it { is_expected.to eq(expected) }

       expected: "https://www.example.com/"
            got: #<Twingly::URL:0x3fe6f02f84a8 https://www.example.com/ >

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"https://www.example.com/"
       +#<Twingly::URL:0x3fe6f02f84a8 https://www.example.com/ >

     # ./spec/lib/twingly/url_spec.rb:210:in `block (5 levels) in <top (required)>'

  4) Twingly::URL.parse with url containing leading and trailing: non-breaking space and carriage-return should eq "https://www.example.com/"
     Failure/Error: it { is_expected.to eq(expected) }

       expected: "https://www.example.com/"
            got: #<Twingly::URL:0x3fe6f0305234 https://www.example.com/ >

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"https://www.example.com/"
       +#<Twingly::URL:0x3fe6f0305234 https://www.example.com/ >

     # ./spec/lib/twingly/url_spec.rb:210:in `block (5 levels) in <top (required)>'
```

spec/lib/twingly/url_spec.rb
```diff
@@ -70,12 +70,30 @@ def valid_urls
...
+  non_breaking_space = "\u00A0"
+  space              = "\u0020"
+
   {
-    "non-breaking space and space" => "\u00A0\u0020",
-    "non-breaking space" => "\u00A0",
-    "non-breaking space, space, non-breaking space" => "\u00A0\u0020\u00A0",
-    "space and non-breaking space" => "\u0020\u00A0",
-    "space, non-breaking space and space" => "\u0020\u00A0\u0020",
+    "non-breaking space and space"                  => [non_breaking_space, space].join,
```
walro: I find the following easier to understand:

```suggestion
    "non-breaking space and space"                  => "#{non_breaking_space}#{space}"
```

(and for the rest)

spec/lib/twingly/url_spec.rb
```diff
@@ -70,12 +70,30 @@ def valid_urls
 end
 
 def leading_and_trailing_whitespace
```
walro: A future improvement could be to generate all permutations instead of manually listing them. It looks like a tedious task over time

spec/lib/twingly/url_spec.rb
dentarg: Oh no 🙈 I started like that but thought it was too fiddly and was very happy when I came up with the join approach :-)

spec/lib/twingly/url_spec.rb
dentarg: Yeah for sure

spec/lib/twingly/url_spec.rb
walro: Hehe, yes it's prettier, but I found myself pondering "what's the default behavior on join"

dentarg: I know it is still basically the same, but what if we DRY the `#join`, does it help or does it makes it even worse? :-)

```ruby
...  
    "carriage-return and line-feed"                 => [carriage_return, line_feed],
    "line-feed and carriage-return"                 => [line_feed, carriage_return],
  }.map { |key, val| [key, val.join] }.to_h

dentarg: I thought it was easier to both read and write the arrays... seeing we will/should generate the combinations if we add more, the "easier to write" might not matter much.

dentarg: > Hehe, yes it's prettier, but I found myself pondering "what's the default behavior of #join"

Hehe, yeah, I've taught myself what it does :-) It is very hand I've come to realize.

walro: We can leave it like this. Let's skip the DRY part IMO :)

PR #145, opened 2020-09-17
Update to Ruby 2.7 and drop support for Ruby 2.4

Pontus4:


twingly-url.gemspec
```diff
@@ -12,10 +12,10 @@ Gem::Specification.new do |s|
   s.summary     = "Ruby library for URL handling"
   s.description = "Twingly URL tools"
   s.license     = "MIT"
-  s.required_ruby_version = "~> 2.4"
+  s.required_ruby_version = "~> 2.5"
 
   s.add_dependency "addressable", "~> 2.6"
-  s.add_dependency "public_suffix", "~> 3.0.1"
+  s.add_dependency "public_suffix", "~> 4.0.3"
```
walro, line 18: maybe we should be a bit more permissive, like:

```suggestion
  s.add_dependency "public_suffix", ">= 3.0.1", "< 5.0", just in case 4.x would be hard to update to in all projects
```

Pontus4: Sure thing!

walro: See cdc0ff9

walro (approved):
LGTM!

walro:
I modified the title slightly, just so that the 2.4 parts gets into the change log.

walro:
I think you can cut this as a 6.0.3 release, doesn't look like anything breaking: https://github.com/twingly/twingly-url/compare/v6.0.2...master

Pontus4:
> I think you can cut this as a 6.0.3 release, doesn't look like anything breaking: [v6.0.2...master](https://github.com/twingly/twingly-url/compare/v6.0.2...master)

6.0.3 it is! (ccc72c931beb7eb65fedfb08297d365038fd3e3e)

PR #148, opened 2020-12-01
Run CI on GitHub actions

walro:
I had to loosen the requirement on Ruby so that we cold run on Ruby head. I split the Travis CI things into two separate workflows as it made sense. One workflow is concerned about running the specs and one is concerned about trying the install phase. The downside here is that we have two workflows and Ruby versions scattered in two files, but I think it's still the lesser of evils.


roback (approved):
> The downside here is that we have two workflows and Ruby versions scattered in two files, but I think it's still the lesser of evils.

Agreed, and right now I don't have any better ideas than that either. So, LGTM :)

PR #149, opened 2020-12-01
Add version constraint for Pry

walro:
Takes care of the following warning when building the gem:

```
WARNING:  open-ended dependency on pry (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
```


Pontus4 (approved):
💯

PR #151, opened 2021-05-27
Run CI on latest Rubies

walro:
Removed Ruby 2.5 from matrix as it went EoL 2021-03-31.


walro:
> and drop support for Ruby 2.5 

Silly me forgot that JRuby is still 2.5 :)

roback (approved):

PR #155, opened 2022-08-19
Allow use of `public_suffix` 5

dentarg:
Changes in `public_suffix` looks fine to me, no breaking API changes what I can see:

* https://github.com/weppos/publicsuffix-ruby/blob/v5.0.0/CHANGELOG.md#500
* https://github.com/weppos/publicsuffix-ruby/compare/v4.0.7...v5.0.0

Addressable 2.8.1 was just released to allow use of public_suffix 5.x:
https://github.com/sporkmonger/addressable/blob/addressable-2.8.1/CHANGELOG.md#addressable-281

### DIFF
--- a/twingly-url.gemspec
+++ b/twingly-url.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.5"
 
   s.add_dependency "addressable", "~> 2.6"
-  s.add_dependency "public_suffix", ">= 3.0.1", "< 5.0"
+  s.add_dependency "public_suffix", ">= 3.0.1", "< 6.0"
 
   s.add_development_dependency "rake", "~> 12"
   s.add_development_dependency "rspec", "~> 3"


dentarg:
Hmm, maybe this part needs some tweaking after https://github.com/sporkmonger/addressable/pull/459

https://github.com/twingly/twingly-url/blob/767807a5a8424c7a31fd1ef172ded0ce3f117a75/lib/twingly/url.rb#L94-L104

Haven't looked into it, I'll leave that to the real maintainers :-)

Pontus4:
> Hmm, maybe this part needs some tweaking after [sporkmonger/addressable#459](https://github.com/sporkmonger/addressable/pull/459)
> 
> https://github.com/twingly/twingly-url/blob/767807a5a8424c7a31fd1ef172ded0ce3f117a75/lib/twingly/url.rb#L94-L104
> 

Maybe this method isn't needed anymore after the updated requirements in [5ae3a68](https://github.com/twingly/twingly-url/pull/155/commits/5ae3a6838cf1202d1969568773f06681342dc261), but as I understand it https://github.com/sporkmonger/addressable/issues/224 isn't completely solved yet. We don't have any URLs in the specs that fails on the `normalize` step, but that doesn't say that it isn't possible :)

spec/lib/twingly/url_spec.rb
```diff
@@ -66,6 +65,7 @@ def valid_urls
     "https://www.foo.ایران.ir/bar",
     "https://www.foo.xn--mgba3a4f16a.ir/bar",
     "http://AcinusFallumTrompetumNullunCreditumVisumEstAtCuadLongumEtCefallumEst.com",
+    "http://+%D5d.some_site.net",
```
dentarg, line 68: I was thinking you wanted to keep this as not valid... is it possible to visit it? :)

Pontus4: > I was thinking you wanted to keep this as not valid... is it possible to visit it? :)

I think you're right. I thought of Addressable's and PublicSuffix's parsers as validators but I understand now that that is not the case. 

> What does Arvegods say, the ol' mighty source of all truths?

It seems like it doesn't buy it.

```cmd
~/repos/arvegods echo http://+%D5d.some_site.net | mono ../arvegods/trunk/Deploy/Normalizer/Normalizer.exe
http://+%D5d.some_site.net,http://+%D5d.some_site.net,8750602076988975287
~/repos/arvegods echo +%D5d.some_site.net | mono ../arvegods/trunk/Deploy/Normalizer/Normalizer.exe
+%D5d.some_site.net,+%D5d.some_site.net,12714123896832132254
```

I guess we will have to implement another validator now then. I tested some of the methods suggested in [this SO thread](https://stackoverflow.com/questions/1805761/how-to-check-if-a-url-is-valid) but they all find it valid.

```ruby
[1] pry(main)> "http://+%D5d.some_site.net" =~ /\A#{URI::regexp}\z/
=> 0
[2] pry(main)> "http://+%D5d.some_site.net" =~ /\A#{URI::regexp(['http', 'https'])}\z/
=> 0
[3] pry(main)> "http://+%D5d.some_site.net" =~ URI::DEFAULT_PARSER.regexp[:ABS_URI]
=> 0
[4] pry(main)> URI.parse("http://+%D5d.some_site.net").kind_of?(URI::HTTP)
=> true
```

dentarg: I think you want to enforce this for hostnames (`host` minus `tld` as the public suffix list will take care of any invalid TLDs, like `"http://some_site.net%C2"` you have above), from https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax,_internationalization

> Although no technical limitation exists to prevent domain name labels using any character which is representable by an octet, hostnames use a preferred format and character set. The characters allowed in labels are a subset of the [ASCII](https://en.wikipedia.org/wiki/ASCII) character set, consisting of characters a through z, A through Z, digits 0 through 9, and hyphen. This rule is known as the LDH rule (letters, digits, hyphen). Domain names are interpreted in case-independent manner.[[25]](https://en.wikipedia.org/wiki/Domain_Name_System#cite_note-25) Labels may not start or end with a hyphen.

Pontus4: I tested native IDNA (as described in the [Addressable README](https://github.com/sporkmonger/addressable/#install)). This takes care of `http://+%D5d.some_site.net` and marks it as invalid, but it also translates `straße.de` into `strasse.de` instead of `xn--strae-oqa.de` ([IDNA 2003 behaviour instead of IDNA 2008](https://curl.se/docs/CVE-2016-8625.html))... I guess this is related to https://github.com/sporkmonger/addressable/issues/408.

Not sure if we can do anything about this on our side.

dentarg: You can contribute that PR to Addressable :-)

dentarg: I've implemented https://github.com/twingly/twingly-url/pull/155/files#r965214795 in https://github.com/twingly/twingly-url/pull/158

I think that should be merged and 5ae3a6838cf1202d1969568773f06681342dc261 removed from this branch (I can do that once #158 has been merged)

Pontus4: #158 has been merged now :)

dentarg: Done :)

walro (approved):

Pontus4 (approved):
🎉

PR #156, opened 2022-08-19
CI housekeeping

dentarg:
Use latest checkout action: https://github.com/actions/checkout/blob/main/CHANGELOG.md

Use Ubuntu 20.04 as 18.04 is going away soon: https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
Chrizpy, line 9: Might as well hop on Jammie immediately 😄 

```suggestion
    runs-on: ubuntu-22.04
```

Chrizpy (approved):
LGTM!

PR #157, opened 2022-08-23
Require Ruby >= 2.6

Pontus4:
Both Ruby 2.5 and 2.6 have reached EOL, but JRuby 9.3 still requires Ruby 2.6, which is why we have to allow it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - { ruby: '2.6' }
           - { ruby: '2.7' }
           - { ruby: '3.0' }
           - { ruby: '3.1' }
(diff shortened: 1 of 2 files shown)


roback (approved):

Chrizpy (approved):

Pontus4:
JRuby 9.3.x is compatible with Ruby 2.6.x, which is why the check fails. I guess we have to lock the Ruby version to `>= 2.6` instead then...

Chrizpy:
> JRuby 9.3.x is compatible with Ruby 2.6.x, which is why the check fails. I guess we have to lock the Ruby version to `>= 2.6` instead then...

Yeah, looking more closely now it seems to me that the latest `JRuby` version only supports Ruby compatibility up to `2.6.x`. 

They have a meta issue [issue](https://github.com/jruby/jruby/issues/6464) since 2020 that collects the efforts for an upcoming `2.7` (maybe also beyond) support. Looks like these changes are coming in into `JRuby-9.4.x.x` That is also why the `jruby-head` in the CI passes, since it uses `jruby 9.4.0.0-SNAPSHOT (3.1.0)` (from CI log).

Chrizpy (approved):

roback (approved):
🟢

PR #158, opened 2022-10-05
Validate the normalized hostname

### LINKED ISSUE #62: Bug in normalized_host in Addressable (ArgumentError: invalid byte sequence in UTF-8)

dentarg:
According to the "preferred format" used by DNS.

See https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax,_internationalization

Moves one invalid URL to the set of invalid URLs (if you enter http://www..twingly..com/ in the address bar in Chrome, it does a search, doesn't try to visit any site).

Close #62

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -22,6 +22,8 @@ module Twingly
       Addressable::URI::InvalidURIError,
       PublicSuffix::DomainInvalid,
     ].freeze
+    DOT = "."
+    HYPHEN = "-"
     CARRIAGE_RETURN = "\u000D"
     LINE_FEED = "\u000A"
     NBSP = "\u00A0"
@@ -34,16 +36,20 @@ module Twingly
     ].join.freeze
     LEADING_AND_TRAILING_WHITESPACE =
       /\A[#{WHITESPACE_CHARS}]+|[#{WHITESPACE_CHARS}]+\z/.freeze
+    LETTERS_DIGITS_HYPHEN = /\A[a-zA-Z0-9-]+\z/.freeze
 
     private_constant :ACCEPTED_SCHEMES
     private_constant :CUSTOM_PSL
     private_constant :STARTS_WITH_WWW
     private_constant :ENDS_WITH_SLASH
     private_constant :ERRORS_TO_EXTEND
+    private_constant :DOT
+    private_constant :HYPHEN
     private_constant :NBSP
     private_constant :SPACE
     private_constant :WHITESPACE_CHARS
     private_constant :LEADING_AND_TRAILING_WHITESPACE
+    private_constant :LETTERS_DIGITS_HYPHEN
 
     class << self
       def parse(potential_url)
@@ -91,10 +97,9 @@ module Twingly
         input.gsub(LEADING_AND_TRAILING_WHITESPACE, "")
       end
 
-      # Workaround for the following bug in addressable:
-      # https://github.com/sporkmonger/addressable/issues/224
       def try_addressable_normalize(addressable_uri)
-        addressable_uri.normalize
+        ascii_host = addressable_uri.normalize.host
+        raise Twingly::URL::Error::ParseError unless valid_hostname?(ascii_host)
       rescue ArgumentError => error
         if error.message.include?("invalid byte sequence in UTF-8")
           raise Twingly::URL::Error::ParseError
@@ -103,11 +108,27 @@ module Twingly
         raise
       end
 
+      def valid_hostname?(hostname)
+        # No need to check the TLD, the public suffix list does that
+        labels = hostname.split(DOT)[0...-1].map(&:to_s)
+
+        labels.all? { |label| valid_label?(label) }
+      end
+
+      def valid_label?(label)
+        return false if label.start_with?(HYPHEN)
+        return false if label.end_with?(HYPHEN)
+
+        label.match?(LETTERS_DIGITS_HYPHEN)
+      end
+
       private :new
       private :internal_parse
       private :clean_input
       private :strip_whitespace
       private :try_addressable_normalize
+      private :valid_hostname?
+      private :valid_label?
     end
 
     def initialize(addressable_uri, public_suffix_domain)
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -39,6 +39,7 @@ def invalid_urls
     "http://.gl/xxx",
     "http://.twingly.com/",
     "http://www.twingly.",
+    "http://www..twingly..com/",
 
     # Test that we can handle upstream bug in Addressable, references:
     # https://github.com/twingly/twingly-url/issues/62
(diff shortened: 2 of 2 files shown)


lib/twingly/url.rb
```diff
@@ -22,6 +22,8 @@ class URL
       Addressable::URI::InvalidURIError,
       PublicSuffix::DomainInvalid,
     ].freeze
+    DOT = "."
+    HYPEN = "-"
```
walro, line 26: should these be marked as private constants too? can't spot usage outside of this file anyway

lib/twingly/url.rb
```diff
@@ -103,11 +108,26 @@ def try_addressable_normalize(addressable_uri)
         raise
       end
 
+      def valid_hostname?(domain)
```
walro, line 111: nitpick: method implies we're validating a hostname but we pass a `domain` variable, ignore me If I don't get the details 👼

spec/lib/twingly/url_spec.rb
```diff
@@ -564,12 +565,6 @@ def leading_and_trailing_whitespace
       it { is_expected.to eq(expected) }
     end
 
-    context "oddly enough, does not alter URLs with consecutive dots" do
```
walro, line 567: hehe, the quirk with `twingly-url` is that is was intentionally meant to mimic the behavior of the good ol' .NET implementation, the change here seems to make the two implementation diverge as I gather, not sure it matters anymore though 🤷

lib/twingly/url.rb
```diff
@@ -103,11 +108,26 @@ def try_addressable_normalize(addressable_uri)
...
+        # No need to check the TLD, the public suffix list does that
+        labels = domain.split(DOT)[..-2].map(&:to_s)
+
+        valid = labels.each do |label|
+          break false if label.start_with?(HYPEN)
+          break false if label.end_with?(HYPEN)
+          break false unless label.match?(LETTERS_DIGITS_HYPHEN)
+          true
+        end
```
walro, line 120: I think you might want `break true` on the last line, or switch to using something like `#all?` instead. Note the subtle difference, in particular for 10 and 12:

```ruby
[10] pry(main)> [1,2,3].each { |num| break true }
=> true
[11] pry(main)> [1,2,3].each { |num| break false }
=> false
[12] pry(main)> [1,2,3].each { |num| true }
=> [1, 2, 3]
```

lib/twingly/url.rb
dentarg: Yes :)

lib/twingly/url.rb
dentarg: Should be called `hostname` here (but we pass `public_suffix_domain.domain` but not necessary to reflect that here)

lib/twingly/url.rb
dentarg: True, but I don't think `break true` is good either, need to check all labels

I wanted to stop checking labels when one was found to be not valid

lib/twingly/url.rb
```diff
@@ -103,11 +108,26 @@ def try_addressable_normalize(addressable_uri)
         raise
       end
 
+      def valid_hostname?(domain)
+        # No need to check the TLD, the public suffix list does that
+        labels = domain.split(DOT)[..-2].map(&:to_s)
```
Pontus4, line 113: I prefer this, but it's highly subjective :)

```suggestion
        labels = domain.split(DOT)[...-1].map(&:to_s)
```

lib/twingly/url.rb
Pontus4: I guess the  final `true` isn't actually needed at all? I don't mind it though, as it could improve readability.

I like @walro's suggestion with `#all?` however!

```suggestion
        valid = labels.all? do |label|
          label.match?(LETTERS_DIGITS_HYPHEN) &&
            !label.start_with?(HYPEN) &&
            !label.end_with?(HYPEN)
        end
```

dentarg: Yes will re-work this

lib/twingly/url.rb
```diff
@@ -103,11 +108,27 @@ def try_addressable_normalize(addressable_uri)
         raise
       end
 
+      def valid_hostname?(hostname)
+        # No need to check the TLD, the public suffix list does that
+        labels = hostname.split(DOT)[0...-1].map(&:to_s)
```
dentarg, line 113: Need the `0` for JRuby 9.3: https://github.com/twingly/twingly-url/actions/runs/3192326669/jobs/5209668938#step:4:30

dentarg:
This is ready now

Pontus4 (approved):
Noticed one tiny thing, then this LGTM!

Pontus4 (approved):
Noticed one tiny thing, then this LGTM!

spec/lib/twingly/url_spec.rb
walro: Just mentioning that normalization of links is/used to be a thing, in those cases it didn't necessarily mean that they'd have to be "visitable".

Chrizpy (approved):

dentarg:
:shipit:

PR #160, opened 2022-10-29
Add missing nil guard in `valid_hostname?`

dentarg:
Regression from https://github.com/twingly/twingly-url/pull/158

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -109,6 +109,8 @@ module Twingly
       end
 
       def valid_hostname?(hostname)
+        return false if hostname.nil?
+
         # No need to check the TLD, the public suffix list does that
         labels = hostname.split(DOT)[0...-1].map(&:to_s)
 
--- a/spec/lib/twingly/url_spec.rb
+++ b/spec/lib/twingly/url_spec.rb
@@ -12,6 +12,7 @@ def invalid_urls
     "http:",
     "htttp",
     "http://",
+    "http:X",
     "a",
     "1",
     "?",


lib/twingly/url.rb
```diff
@@ -109,6 +109,8 @@ def try_addressable_normalize(addressable_uri)
       end
 
       def valid_hostname?(hostname)
+        return false if hostname.nil?
+
         # No need to check the TLD, the public suffix list does that
         labels = hostname.split(DOT)[0...-1].map(&:to_s)
```
walro, line 115: Just some alternatives. I think current is just as good as them, but options are always fun!

To avoid work later on, not sure if `hostname` can ever be `""` though:

```suggestion
        return false if hostname.to_s.empty?

        # No need to check the TLD, the public suffix list does that
        labels = hostname.split(DOT)[0...-1].map(&:to_s)
```

or even

```suggestion
        # No need to check the TLD, the public suffix list does that
        labels = hostname.to_s.split(DOT)[0...-1].map(&:to_s)
```

I used to like the, latter, "benign value approach", but not sure I do these days :)

walro (approved):

lib/twingly/url.rb
dentarg: I thought of doing `to_s` (and guard against the empty string) but opted not to because the code already handles empty string. Sure, avoiding unnecessary work could be good. I will leave it to someone else to [profile](https://github.com/twingly/twingly-url#profiling) if they want. :)

Pontus4: We use to use guard clauses in situations like these. I'm fine with both the committed solution and the first suggestion here, so let's just go with what's already committed :D

Pontus4 (approved):
👍

dentarg:
A release with this fix would be great :)

Pontus4:
> A release with this fix would be great :)

[v7.0.1](https://github.com/twingly/twingly-url/releases/tag/v7.0.1) 🎉

PR #163, opened 2025-09-12
Don't redo the parsing when normalizing the url.

bjornkellner:
The parsing isn't redone when calling normalized. 
I don't want to make any big changes in the this gem right now. But this is minimal change that will do a significant speedup with what I consider a low risk of introducing changed behavior/bugs.

### DIFF
--- a/lib/twingly/url.rb
+++ b/lib/twingly/url.rb
@@ -73,11 +73,7 @@ module Twingly
         try_addressable_normalize(addressable_uri)
 
         host = addressable_uri.host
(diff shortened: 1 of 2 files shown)


roback (commented):
Had one suggestion.

One "disadvantage" with this change, is that when we now call `#normalize`, we won't rescue potential `PublicSuffix::DomainInvalid` errors (since we don't call the `.parse` method inside the `#normalize` method anymore, which rescues these errors).

However, I'm not sure if it's possible for that error to get raised in this case, considering the non-normalized URL has already passed through the parse method before.

PR #167, opened 2026-09-01
Add optional percent-encoding to Twingly::URL::Extended

Eric-Twingly:
Adds an opt-in `percent_encode` keyword to `normalize_and_calculate_urlhash` that canonicalizes the URL with `Addressable::URI#normalize` first. 

Identical to encoding applied before fetching in https://github.com/twingly/phelps/blob/master/lib/fetch/job.rb#L18-#L23.


lib/twingly/url/extended.rb
```diff
@@ -149,8 +151,15 @@ def self.calculate_urlhash(url)
         Twingly::URL::Hasher.documentdb_hash(url).to_s
       end
 
+      def self.canonicalize_percent_encoding(url)
+        Addressable::URI.parse(url).normalize.to_s
```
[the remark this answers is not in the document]
Eric-Twingly: This intentionally mirrors the percent encoding from [Phelps](https://github.com/twingly/phelps/blob/master/lib/fetch/job.rb#L19).

lib/twingly/url/extended.rb
```diff
@@ -117,9 +117,11 @@ def normalized_fragment
         nil
       end
 
-      def self.normalize_and_calculate_urlhash(url)
+      def self.normalize_and_calculate_urlhash(url, percent_encode: false)
```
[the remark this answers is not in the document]
Eric-Twingly: Added to documentation in ddd839e.

bjornkellner, line 120: A bit missleading name, maybe
```suggestion
      def self.normalize_and_calculate_urlhash(url, addressable_normalize: false)
```

lib/twingly/url/extended.rb
```diff
@@ -117,9 +117,11 @@ def normalized_fragment
...
         return empty_result if url.to_s.strip.empty?
 
+        url = canonicalize_percent_encoding(url.to_s) if percent_encode
```
bjornkellner, line 123: We don't want to convert the url to string and then parse it again if it can be avoided.

More generally  I think this class should expose 3 urls. 
* the original_url  
* a "callable url" - a url normalized so that it can be used in phelps.
* a fully normalized url. (to use for hashing and url comparison)

I would  create a function similar to current `normalized`  called `addressable_normalized` or maybe `callable` 

and a private class function  `normalized_addressable_url` 


```diff
+def addressable_normalized 
+       normalized_url = normalized_addressable_url
+       public_suffix_domain = get_public_suffix_domain(normalized_url.host)
+       self.class.send(:new, normalized_url, public_suffix_domain)
+end
+private 
+
+def normalized_addressable_url
+      addressable_uri.dup.normalized
+end

```
If `normalized` is updated ass suggested in https://github.com/twingly/twingly-url/pull/167/changes#r3913551400 this function could look like:

```diff
- def self.normalize_and_calculate_urlhash(url, percent_encode: false)
+ def self.normalize_and_calculate_urlhash(url, addressable_normalize: false)
-     url = canonicalize_percent_encoding(url.to_s) if percent_encode
      twingly_url = if url.is_a?(Extended)
                        url
                      else
                        Extended.parse(url)
                      end

        return empty_result unless twingly_url.valid?

        original_url                  = twingly_url.original_url_without_blacklisted_parameters
-       normalized_url                = twingly_url.normalized.to_s
+       normalized_url                = twingly_url.normalized(addressable_normalize:).to_s
        normalized_url_without_scheme = remove_scheme(normalized_url)
        urlhash                       = calculate_urlhash(normalized_url_without_scheme)
        legacy_urlhash                = calculate_urlhash(normalized_url)

        HashResult.new(original_url, normalized_url_without_scheme, urlhash, legacy_urlhash)
      end
```

Depending on if what we want  HashResult.url to be 
 this could also be changed:
```diff
-original_url                  = twingly_url.original_url_without_blacklisted_parameters
+original_url                  = twingly_url.addressable_normalized.original_url_without_blacklisted_parameters
```

lib/twingly/url/extended.rb
bjornkellner, line 47: I think `normalized` also should apply all normalizations. See 
https://github.com/twingly/twingly-url/pull/167/changes#r3913359676 

```diff
-  def normalized
+ def  normalized(addressable_normalize: false)
       normalized_url = addressable_uri.dup

        normalized_url.scheme       = normalized_scheme
        normalized_url.host         = normalized_host
        normalized_url.path         = normalized_path
        normalized_url.query_values = normalized_query
        normalized_url.fragment     = normalized_fragment
      
        public_suffix_domain = get_public_suffix_domain(normalized_url.host)
+       normalized_url =  normalize_addressable_url(normalized_url) if addressable_normalize
        self.class.send(:new, normalized_url, public_suffix_domain)

```

bjornkellner (commented):
This class and the underlying gem have become a bit of a mess 😅 .
I think a bigger rework is needed later. Good job with this though.

I think that `normalized` also should have the updated normalization, not only `normalize_and_calculate_urlhash`

The comment's are a suggested solution.

PR #169, opened 2026-09-09
Add optional addressable-normalize to Twingly::URL::Extended

Eric-Twingly:
Rename of https://github.com/twingly/twingly-url/pull/167.

Adds an opt-in `addressable normalize` to extended normalization.


Eric-Twingly:
@bjornkellner Renaming the branch auto closed the original, tried to address your comments from https://github.com/twingly/twingly-url/pull/167 in 6653091.

lib/twingly/url/extended.rb
```diff
@@ -13,7 +13,7 @@ class Extended < Twingly::URL
       # This way the urlhashes here will be the same as the one we have in Elasticsearch
       PROTOCOL_EXPRESSION = /\Ahttps?:/i
 
-      HashResult = Struct.new(:url, :normalized_url, :urlhash, :legacy_urlhash)
+      HashResult = Struct.new(:url, :addressable_normalized_url, :normalized_url, :urlhash, :legacy_urlhash)
```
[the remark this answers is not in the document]
Eric-Twingly: This establishes the contract since no-one is currently using the extended class yet.

lib/twingly/url/extended.rb
```diff
@@ -44,7 +44,9 @@ class Extended < Twingly::URL
 
       # Taken from twingly-url, with the addition of normalizing the query and fragment components
       # See https://github.com/twingly/twingly-url/blob/e20f5fce077d93e89ef8520961be453c90cfec8c/lib/twingly/url.rb#L185-L193
-      def normalized # rubocop:disable Metrics/AbcSize
+      def normalized(addressable_normalize: false) # rubocop:disable Metrics/AbcSize
+        return addressable_normalized.normalized if addressable_normalize
```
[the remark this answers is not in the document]
Eric-Twingly: Fixed in 9fe8612.

lib/twingly/url/extended.rb
```diff
@@ -44,7 +44,9 @@ class Extended < Twingly::URL
...
+
         normalized_url = addressable_uri.dup
```
bjornkellner, line 50: I think this is easier to understand,  and skips the extra call to get_public_suffix_domain, self.class.send.
But no strong opinion.


```suggestion
        normalized_url = addressable_uri.dup
        normalized_url =  normalized_url.normalize if addressable_normalize

```

lib/twingly/url/extended.rb
```diff
@@ -128,17 +140,22 @@ def self.normalize_and_calculate_urlhash(url)
 
         return empty_result unless twingly_url.valid?
 
-        original_url                  = twingly_url.original_url_without_blacklisted_parameters
+        original_url               = twingly_url.original_url_without_blacklisted_parameters
+        addressable_normalized     = twingly_url.addressable_normalized
```
bjornkellner, line 144: I consider `addressable_normalized_url`  the new "standard url" that we will use and pass through our system, i think it should have the blacklisted parameters removed. 

Same in the [addressable_normalized](https://github.com/twingly/twingly-url/pull/169/changes#diff-8ca2862a0ee4d09234d8803a0132f21ed783c089c5d8d98d23dd281816907e59R64) function. Maybe add it there, even though the name get's a bit misguiding.

lib/twingly/url/extended.rb
Eric-Twingly: If you prefer this method I'll happily implement it, but I believe we'd have to resolve an issue with your proposed implementation. Let me know your preferred solution. 

The `addressable_uri` is initialized in [Twingly::URL#initialize](https://github.com/twingly/twingly-url/blob/v7.1.1/addressable-normalize/lib/twingly/url.rb#L136-L139). Your proposed change would set the `normalized_url` to the `addressable normalized URL`, however the subsequent normalization steps would still use the original `addressable_uri` set in the original initialize. For example, [normalized_path](https://github.com/twingly/twingly-url/blob/v7.1.1/addressable-normalize/lib/twingly/url/extended.rb#L97-L101) would override the `addressable_normalized` `path` with the `addressable_uri` `path`.

lib/twingly/url/extended.rb
Eric-Twingly: Added blacklisted parameter stripping to the addressable normalized URL in 10c9129. I opted to not include the blacklisted parameter stripping in the addressable_normalized function because it would be misleading, like you suggested.

lib/twingly/url/extended.rb
bjornkellner: A yes. Never mid. Don't know what I was thinking.

bjornkellner (approved):
Nice job!
maybe update [version](https://github.com/twingly/twingly-url/blame/aa291e369b731f593f5492d6adce2a5c2fa9ba45/lib/twingly/version.rb) ?

